PR #20, opened 2025-05-03
Generic support for upgrades and for action module as v1.0.0

a-ok123:
Added generic support for upgrades and for action module as v1.0.0


akobrin1 (approved):
Should we use versioning for action module from the beginning? (https://docs.cosmos.network/main/build/architecture/adr-054-semver-compatible-modules) - like "/x/action/v1" instead of "x/action". That will make so much easier to maintain multiple versions of action module - in case we introduce some breaking changes like changing field names/types, params. etc.

mateeullahmalik (approved):

PR #58, opened 2025-10-16
Refund Expired Action Price 

akobrin1:
- refund expired action price to the action's creator
- added unit and integration tests


x/action/v1/module/module.go
```diff
@@ -31,8 +31,9 @@ var (
 	_ appmodule.AppModule       = (*AppModule)(nil)
 	_ appmodule.HasBeginBlocker = (*AppModule)(nil)
 	_ appmodule.HasEndBlocker   = (*AppModule)(nil)
-// TODO: Uncomment the following line when the IBC module is implemented
-//	_ porttypes.IBCModule       = (*IBCModule)(nil)
+
+	// TODO: Uncomment the following line when the IBC module is implemented
+	// _ porttypes.IBCModule       = (*IBCModule)(nil)
```
a-ok123: @akobrin1  Don't we need to uncomment it?

akobrin1: no, IBC for action module is not implemented yet (Ignite CLI generated it - but I've set it aside for now, we need to properly integrate it)

PR #105, opened 2026-03-10
test: add empty active set deadlock and bootstrap tests for audit module

mateeullahmalik:
## Summary

Add unit and system tests that reproduce the testnet deadlock where all supernodes are POSTPONED and the epoch anchor has an empty active set. Without active probers, peer observations cannot be generated, making audit recovery impossible.

### Unit tests (`x/audit/v1/keeper/enforcement_empty_active_set_test.go`)

- **`TestEnforceEpochEnd_EmptyActiveSet_PostponedCannotRecover`** — Proves that compliant host-only reports from POSTPONED SNs are insufficient for recovery when no peer observations exist.
- **`TestEnforceEpochEnd_LegacyRecoveredSN_SurvivesWithReport`** — Proves that SNs recovered to ACTIVE mid-epoch (via legacy metrics) with audit reports survive the EndBlocker enforcement.

### System tests (`tests/systemtests/audit_empty_active_set_bootstrap_test.go`)

- **`TestAuditEmptyActiveSetDeadlock_HostOnlyReportsCannotRecover`** — Full-chain E2E: registers SNs, misses epoch 0, submits host-only reports for 3 epochs — all remain POSTPONED throughout.
- **`TestAuditEmptyActiveSetBootstrap_LegacyMetricsBreaksDeadlock`** — Full-chain E2E: same deadlock setup, then legacy metrics recovery breaks the deadlock — SNs survive enforcement and remain ACTIVE.

## Context

On testnet (`lumera-testnet-2`), all supernodes running `v2.4.5-testnet` were POSTPONED before the audit module activated (legacy staleness handler). The audit module's recovery requires peer observations from active probers, but the active set has been empty since epoch 1. Fix applied in [supernode#275](https://github.com/LumeraProtocol/supernode/pull/275).

These tests ensure the edge case is covered for future upgrades.

### DIFF
--- a/tests/systemtests/audit_peer_ports_enforcement_test.go
+++ b/tests/systemtests/audit_peer_ports_enforcement_test.go
@@ -4,10 +4,19 @@ package system
 
 import (
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	"github.com/tidwall/sjson"
 )
+
+func awaitAtLeastHeightWithSlackPeerPorts(t *testing.T, height int64) {
+	t.Helper()
+	if sut.currentHeight >= height {
+		return
+	}
+	sut.AwaitBlockHeight(t, height, 45*time.Second)
+}
 
 func TestAuditPeerPortsUnanimousClosedPostponesAfterConsecutiveWindows(t *testing.T) {
 	const (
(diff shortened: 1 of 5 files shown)


tests/systemtests/audit_peer_ports_enforcement_test.go
```diff
@@ -4,11 +4,20 @@ package system
...
 	"github.com/tidwall/sjson"
 )
 
+func awaitAtLeastHeightWithSlackPeerPorts(t *testing.T, height int64) {
+	t.Helper()
+	if sut.currentHeight >= height {
+		return
+	}
+	sut.AwaitBlockHeight(t, height, 45*time.Second)
+}
```
ghost, line 19: Minor: `awaitAtLeastHeightWithSlackPeerPorts` is identical to `awaitAtLeastHeightWithSlack` defined in `audit_empty_active_set_bootstrap_test.go`. Since both are in the same package, you could drop this copy and call the existing one directly, or move the shared version into `audit_test_helpers_test.go` (replacing the current `awaitAtLeastHeight` or alongside it).

<sub>[Fix it with Roo Code](https://app.roocode.com/fix-issue/?repo=LumeraProtocol/lumera&prNumber=105&fixId=6b799fb7b9) or mention @roomote and request a fix.</sub>

j-rafique (approved):

PR #117, opened 2026-04-20
audit: add LEP-6 foundation storage-truth scaffolding

j-rafique:
## Summary
Introduces LEP-6 foundation scaffolding in `lumera` as an additive, behavior-neutral change set.

This PR establishes the on-chain protocol/state/query surface needed for later LEP-6 milestones, while intentionally keeping current runtime behavior unchanged.

## Scope
- Repo: `lumera` only
- Branch: `LEP-6-foundation`
- No enforcement activation, no scoring behavior switch, no heal/recheck runtime execution yet

### 1) Proto/API foundation (additive)
- Added LEP-6 storage-truth models and enums in audit proto
- Added `storage_proof_results` to `EpochReport` (compat field retained)
- Added new tx message surfaces:
  - `SubmitStorageRecheckEvidence`
  - `ClaimHealComplete`
  - `SubmitHealVerification`
- Added LEP-6 query surfaces for:
  - node suspicion
  - reporter reliability
  - ticket deterioration
  - heal-op (single and list)
- Added LEP-6 params and genesis state surfaces, including heal-op id counter

### 2) State + keeper wiring
- Added typed state records and CRUD helpers for:
  - Node suspicion state
  - Reporter reliability state
  - Ticket deterioration state
  - Heal-op state + status
- Added store prefixes and deterministic key/index wiring
- Added query server wiring for all new LEP-6 query endpoints
- Added genesis import/export support for all new LEP-6 state collections

### 3) Params/defaults
- Added defaults and validation for new LEP-6 params
- Defaults are no-op/safe to preserve current chain behavior

### 4) Message handlers (foundation-only)
- Registered new message interfaces
- Added behavior-neutral placeholder handlers returning `ErrNotImplemented` (runtime logic deferred to later PRs)

## Follow-up Fixes Applied After Review
Addressed PR review comments:
- Nil request validation in placeholder msg handlers now returns `ErrInvalidRequest` (instead of `ErrInvalidSigner`)
- Removed unintended self-referential `lumera` checksum entry from `devnet/go.sum`

## Compatibility / Risk
- Additive and non-breaking
- Existing active flows remain unchanged (`MsgSubmitEpochReport`, existing enforcement/evidence paths)

## Testing
- Added/updated keeper, query, params, genesis, and simulation-oriented coverage for introduced foundation surfaces
- Verified


x/audit/v1/keeper/msg_storage_truth_placeholders.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+
+	errorsmod "cosmossdk.io/errors"
+
+	"github.com/LumeraProtocol/lumera/x/audit/v1/types"
+)
+
+func (m msgServer) SubmitStorageRecheckEvidence(_ context.Context, req *types.MsgSubmitStorageRecheckEvidence) (*types.MsgSubmitStorageRecheckEvidenceResponse, error) {
+	if req == nil {
+		return nil, errorsmod.Wrap(types.ErrInvalidSigner, "empty request")
```
ghost: All three placeholder handlers wrap `ErrInvalidSigner` (error code 1) for nil-request and empty-creator checks. For the nil-request case this is semantically misleading -- a nil request is an invalid argument, not an invalid signer. Clients that match on error codes will misclassify the failure. Consider using a dedicated `ErrInvalidRequest` or `sdkerrors.ErrInvalidRequest` for the nil check, and reserving `ErrInvalidSigner` for the creator-specific validation.

<sub>[Fix it with Roo Code](https://app.roocode.com/fix-issue/?repo=LumeraProtocol/lumera&prNumber=117&fixId=3f84c4ae29) or mention @roomote and request a fix.</sub>

devnet/go.sum
```diff
@@ -109,6 +109,7 @@ github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.50
 github.com/GoogleCloudPlatform/opentelemetry-operations-go/internal/resourcemapping v0.50.0 h1:ig/FpDD2JofP/NExKQUbn7uOSZzJAQqogfqluZK4ed4=
 github.com/GoogleCloudPlatform/opentelemetry-operations-go/internal/resourcemapping v0.50.0/go.mod h1:otE2jQekW/PqXk1Awf5lmfokJx4uwuqcj1ab5SpGeW0=
 github.com/Knetic/govaluate v3.0.1-0.20171022003610-9aa49832a739+incompatible/go.mod h1:r7JcOSlj0wfOMncg0iLm8Leh48TZaKVeNIfJntJ2wa0=
+github.com/LumeraProtocol/lumera v1.10.0/go.mod h1:p2sZZG3bLzSBdaW883qjuU3DXXY4NJzTTwLywr8uI0w=
```
ghost: This adds a `github.com/LumeraProtocol/lumera v1.10.0/go.mod` entry to `devnet/go.sum`, which is a self-referential dependency (the devnet module depending on its own parent repo at a tagged version). This is likely an artifact of running `go mod tidy` with a stale module cache or a `replace` directive that was temporarily removed. If the devnet module uses a `replace` directive to point at the local `../` path, this checksum line is dead weight and could cause confusion if the tagged version diverges from the local code. Worth confirming this entry is intentional.

<sub>[Fix it with Roo Code](https://app.roocode.com/fix-issue/?repo=LumeraProtocol/lumera&prNumber=117&fixId=62d6261049) or mention @roomote and request a fix.</sub>

mateeullahmalik:
# Production-gate review by Zee — 3 findings

**Methodology**: full file-by-file read of every non-generated changed file in this PR's diff (pr-117 vs its base branch), cross-checked against:
- LEP-6 spec (Notion source of truth)
- `invariant-first-coding` skill (write-path enumeration, sibling symmetry, single source of truth, post-fix re-audit)
- Cosmos SDK consensus discipline (no float, no map iteration, bounded EndBlock, genesis round-trip, errorsmod wrapping)

**Status legend**: each finding's status is computed at the **PR #122 stack-tip** (consensus-gap-fixes commit `a51c439`), so 'FIXED' means a downstream PR in the stack already addresses it; 'OPEN' means it is still present at the tip and must be fixed before merge / Phase-2 activation. Severity rubric in the charter (`~/work/lep6-review/ctx/charter.md`): CRITICAL = consensus halt / state corruption / non-determinism in ABCI; HIGH = spec mismatch with economic impact, missing genesis round-trip, replay enabler; MEDIUM = invariant asymmetry without immediate exploit, unbounded loop with practical bound, missing param validation.

**Severity breakdown:** HIGH=2, MEDIUM=1

---

## 117-F1 — Reporter-reliability threshold validation inequality reversed; spec values rejected
- **Severity**: `HIGH`
- **File**: `x/audit/v1/types/params.go`
- **Lines**: 536-538; defaults 128-129
- **Status at PR #122 tip**: FIXED in PR #122 (defaults now 20/90 with `>= 0` lower bound and `<=` ordering against degraded threshold). Confirm before merge train collapses.
- **What**: `Validate()` requires `LowTrustThreshold >= IneligibleThreshold` and defaults are negative (-20, -50). LEP-6 §15.4 says R is non-negative with bands `<20 / 20-49 / 50-89 / 90+`, so spec-correct gov updates (LowTrust=20, Ineligible=90) get rejected by Validate.

## 117-F2 — TicketDeterioration heal-threshold default off by 2× from spec
- **Severity**: `HIGH`
- **File**: `x/audit/v1/types/params.go`
- **Lines**: 130
- **Status at PR #122 tip**: FIXED in PR #122 (default → 50). Confirm.
- **What**: `DefaultStorageTruthTicketDeteriorationHealThreshold = 100`; LEP-6 §16 fixes heal-candidate band at `D >= 50`. At default a chain would require 2× the spec score before scheduling any heal op.

## 117-F3 — NodeSuspicion postpone default off-spec; single threshold conflates §17 two bands
- **Severity**: `MEDIUM`
- **File**: `x/audit/v1/types/params.go`
- **Lines**: 127
- **Status at PR #122 tip**: PARTIALLY FIXED in PR #122 (default → 90; the second 'strong postpone' band 140 still has no behavioural distinction — see PR #121 Finding 12).
- **What**: Default 100 sits between spec's two bands (90 postpone-candidate, 140 strong-postpone). Single param cannot represent both.

---

This review is posted as a `COMMENT` (not `REQUEST_CHANGES`) so it does not block merge mechanically — but the CRITICAL and HIGH items must be triaged before activation. I'm available to walk through any of these in detail.

— Zee

x/audit/v1/simulation/submit_epoch_report_variance.go
```diff
@@ -34,7 +34,6 @@ func SimulateMsgSubmitEpochReportVariance(k keeper.Keeper) simtypes.Operation {
 			MemUsagePercent:        10 + r.Float64()*20,
 			DiskUsagePercent:       70 + r.Float64()*35, // exercises both sides of 90% threshold
 			FailedActionsCount:     uint32(r.Intn(3)),
```
ghost, line 36: This removes `CascadeKademliaDbBytes` from the simulated `HostReport`, so the epoch-report-variance simulation will always submit `0` for that field. The field is still in the proto and is actively read by `x/supernode/v1/keeper/audit_metrics.go` for Everlight payout weighting. If the removal is intentional (e.g. the field is being deprecated), this is fine; otherwise it silently drops simulation coverage for the payout-weight path.

<sub>[Fix it with Roo Code](https://app.roocode.com/fix-issue/?repo=LumeraProtocol/lumera&prNumber=117&fixId=ccc84836f5) or mention @roomote and request a fix.</sub>

mateeullahmalik (commented):
# Production-gate re-review by Zee — PR #117 (LEP-6 consolidated)

**Tip reviewed:** `868cbc7c32f9bdadd681df79ef84fbee2ed903a3`
**Methodology:** full file-by-file read at tip + cross-validation of all 42 findings from the original 6-PR stack review + new bug hunt against LEP-6 spec §6/§14/§15/§16/§17/§19/§20/§22 + invariant-first-coding discipline (sibling-symmetry, write-path enumeration, single-source-of-truth, determinism scan).

This is a **production-gate** review. Filtered to real bugs, potential bugs, and consensus-affecting concerns. No nits, no style.

---

## TL;DR

- **28 of the 42 prior findings** from the original stack are now **FIXED**. Good progress, especially the CRITICAL `121-F16` (float64 in divergence consensus path) is correctly converted to integer cross-multiply.
- **6 prior findings remain PARTIAL** — the headline is the `122-F2` audit-hook fallback (still risks bricking finalize) and `121-F12` strong-postpone band has a distinct event but no distinct on-chain action.
- **15 NEW findings** discovered, including:
  - **2 HIGH** consensus / state-correctness bugs (genesis omits epoch-scoped fact indexes; expired heal-ops never advance probation).
  - **1 HIGH** **money-flow regression** (silent removal of supernode `RegistrationFeeShareBps` 2% routing in `DistributeFees`).
  - **1 HIGH** uint64 underflow on `WindowStartEpoch`.
  - Several MEDIUM issues around per-result hot-path scans, trust-multiplier scope, and post-heal counter resets.

**Recommendation:** Do **NOT** merge until the HIGH-severity items below (especially **NEW-C-3 fee-share removal**, **NEW-C-1 genesis omission**, **NEW-A-12 underflow**, **NEW-B-1 expired-heal**) are addressed.

---

## CRITICAL / HIGH — must address before merge

### NEW-C-3 (HIGH): Money-flow regression — supernode `RegistrationFeeShareBps` (2%) silently removed from `DistributeFees`

- **File**: `x/action/v1/keeper/action.go` (deleted master:633-650), `x/action/v1/keeper/keeper.go` (removed `rewardDistributionKeeper` field + constructor arg)
- **Evidence**: `git diff master..HEAD` shows the entire reward-distribution routing block removed:

  ```go
  // REMOVED:
  if k.rewardDistributionKeeper != nil {
      rewardDistributionBps := k.rewardDistributionKeeper.GetRegistrationFeeShareBps(ctx)
      if rewardDistributionBps > 0 && fee.Amount.GT(math.ZeroInt()) {
          rewardDistributionAmount := fee.Amount.MulRaw(int64(rewardDistributionBps)).QuoRaw(10000)
          ...
          err := k.bankKeeper.SendCoinsFromModuleToModule(ctx, actiontypes.ModuleName, sntypes.ModuleName, ...)
          ...
          fee.Amount = fee.Amount.Sub(rewardDistributionAmount)
      }
  }
  ```
  And `GetRewardDistributionKeeper`, the `RewardDistributionKeeper` field, and the constructor parameter were all removed.
- **Why it matters**: Master routes 2% of every cascade fee to the supernode reward pool (default `RegistrationFeeShareBps=200`). PR #117 silently reverts that. The `RegistrationFeeShareBps` param still exists in supernode params, the test `x/supernode/v1/keeper/fee_routing_test.go` still references its consumer being the action module, and the supernode keeper still exposes `GetRegistrationFeeShareBps`. The fees that were earmarked for the reward pool will now flow through to the existing supernode-list distribution path.
- **Spec ref**: out-of-scope for LEP-6; this is unrelated to storage-truth and should not be in this PR.
- **Fix**: either (a) restore the integration in `DistributeFees` exactly as in master, or (b) split the removal into its own PR with explicit deprecation of the supernode param and a callout in the PR description.

### NEW-C-1 (HIGH): `ExportGenesis` omits all new `st/...` epoch-scoped audit state — fact indexes, recheck dedup keys, transcripts, failed-heal markers, and reports are silently dropped on chain export

- **File**: `x/audit/v1/keeper/genesis.go:104-163`
- **Spec ref**: §17 / §22 (state model)
- **Evidence**: `ExportGenesis` exports only `Params`, `Evidence + NextEvidenceID`, the three score-state slices, `TicketArtifactCountStates`, `HealOps + NextHealOpId`, and `StorageTruthPostponements`. It does **not** iterate or serialize:
  - `st/rce/` (recheck-evidence dedup keys — replay protection)
  - `st/nf/` (node-failure fact index)
  - `st/rrs/`, `st/rrs-tt/` (reporter result fact indexes — used by `hasIndependentReporterPassInWindow`)
  - `st/spt/`, `st/spt-tbe/` (storage-proof transcript records — used by `hasCleanRecheckInWindow`, `markStorageTruthReporterResultRecheck`, contradiction confirmation)
  - `st/fh/` (failed-heal markers — consumed by `hasStorageTruthFailedHeal` in strong-postpone predicate)
  - `r/`, `ri/`, `hr/`, `sc/` (epoch reports / indices)
- **Why it matters**: After `ExportGenesis | InitGenesis` (state-sync, chain export-then-import upgrade path), every postpone/recovery/contradiction predicate that depends on the fact indexes silently returns empty:
  1. `storageTruthPostponePredicatesMet` (enforcement.go:646-706) returns false ⇒ NO postpone fires until ≥ classAWindow=14 epochs of fresh failures accumulate.
  2. `markStorageTruthReporterResultRecheck` reads `getStorageProofTranscriptRecord` which is now nil ⇒ recheck score updates silently no-op.
  3. `linkStorageTruthRecheckTranscript` 122-F3 collision check is void (no records exist), so post-export all replay protection is gone.
  4. `RecheckEvidence` dedup (`HasRecheckEvidence`) returns false for every replay submitted right after import.
- **Fix**: Add export/import for all `st/*` epoch-scoped indexes plus the `r/` prefix at minimum. At a bare minimum, fail-fast in `ExportGenesis` when chain has any postponed supernodes whose underlying evidence cannot round-trip.

### NEW-A-12 (HIGH): `WindowStartEpoch > currentEpoch` not validated at genesis — uint64 underflow on first update

- **File**: `x/audit/v1/keeper/storage_truth_scoring.go:256` and `:332`; `x/audit/v1/types/genesis_validate.go:7-30` (no `WindowStartEpoch` check)
- **Spec ref**: §14, §15
- **Evidence**:

  ```go
  // updateNodeSuspicionHistoryFields:
  if epochID-state.WindowStartEpoch >= window {       // line 256 — RAW uint64 subtraction
      state.WindowStartEpoch = epochID
      ...
  }
  // applyReporterReliabilityDelta:
  if epochID-state.WindowStartEpoch >= divergenceWindow { // line 332 — same shape
      nextState.WindowStartEpoch = epochID
      ...
  }
  ```
  Both are raw `uint64` subtractions. `ValidateScoreStatesGenesis` only checks `LastUpdatedEpoch <= currentEpoch`; it never inspects `WindowStartEpoch`. Sibling asymmetry: `epochDelta(currentEpoch, lastUpdatedEpoch)` IS uint64-safe (returns 0 on inversion), so `decayTowardZero` is robust — but the window-reset sites at 256 and 332 are not.
- **Why it matters**: Any future state-sync or re-genesis path that brings in a `WindowStartEpoch` from a higher-epoch sibling chain will silently corrupt window-aggregate counters on first update. Compounded with NEW-C-1 above (genesis omission), divergence-input becomes unreliable.
- **Fix**: extend `ValidateScoreStatesGenesis` to reject `s.WindowStartEpoch > currentEpoch` for both `NodeSuspicionStates` and `ReporterReliabilityStates`. Optionally rewrite both window checks to use `epochDelta` for symmetric defense.

### NEW-B-1 (HIGH): EXPIRED heal-ops do not advance probation or apply §20 — heal-spam loop on an assigned-but-silent healer

- **File**: `x/audit/v1/keeper/storage_truth_heal_ops.go:36-48` (`expireStorageTruthHealOpsAtEpochEnd`)
- **Spec ref**: §19 / §20 (intent)
- **Evidence**:

  ```go
  healOp.Status = types.HealOpStatus_HEAL_OP_STATUS_EXPIRED
  healOp.UpdatedHeight = uint64(ctx.BlockHeight())
  if err := k.SetHealOp(ctx, healOp); err != nil { return err }
  ticketState, found := k.GetTicketDeteriorationState(ctx, healOp.TicketId)
  if found && ticketState.ActiveHealOpId == healOp.HealOpId {
      ticketState.ActiveHealOpId = 0
      if err := k.SetTicketDeteriorationState(ctx, ticketState); err != nil { return err }
  }
  ```
  Compare with the FAILED branch (msg_storage_truth.go:407-412) which advances `ProbationUntilEpoch` AND adds +15 to `D`. Sibling asymmetry: an EXPIRED heal — where the assigned healer never even reported — is at least as suspicious as a FAILED heal, yet it incurs no penalty and clears the active marker.
- **Why it matters**: The very next epoch's `scheduleStorageTruthHealOpsAtEpochEnd` will immediately re-schedule the same ticket because `D` is unchanged and probation is not advanced. A misbehaving healer can keep the ticket on the schedule list indefinitely while burning the `MaxSelfHealOpsPerEpoch` budget and never accumulating any penalty.
- **Fix**:

  ```go
  ticketState.ProbationUntilEpoch = epochID + uint64(params.StorageTruthProbationEpochs)
  ticketState.DeteriorationScore = addInt64Saturated(ticketState.DeteriorationScore, 15) // or a smaller "no-show" delta
  ```
  Mirror the failed-branch cooldown.

---

## MEDIUM — should address before merge

### NEW-C-2 (MEDIUM): FinalizeAction audit hook still passes raw 0-count — single-source-of-truth violation of 122-F2 fix

- **File**: `x/action/v1/keeper/action.go:245-261`
- **Evidence**: `Process` and `GetUpdatedMetadata` were patched with `len(RqIdsIds)` fallback for `IndexArtifactCount` / `SymbolArtifactCount`. The FinalizeAction hook itself still does:

  ```go
  k.auditKeeper.SetStorageTruthTicketArtifactCounts(
      ctx, existingAction.ActionID,
      cascadeMeta.GetIndexArtifactCount(),     // raw — could be 0
      cascadeMeta.GetSymbolArtifactCount(),    // raw — could be 0
  )
  ```
  Audit-side `SetStorageTruthTicketArtifactCounts` returns `ErrInvalidStorageProofs` on either count == 0 → finalize tx hard-reverts → ticket can never be settled.
- **Fix**: mirror the fallback inline before calling the audit keeper:

  ```go
  ic := cascadeMeta.GetIndexArtifactCount()
  sc := cascadeMeta.GetSymbolArtifactCount()
  if ic == 0 { ic = uint32(len(cascadeMeta.GetRqIdsIds())) }
  if sc == 0 { sc = uint32(len(cascadeMeta.GetRqIdsIds())) }
  ```

### NEW-A-11 residue (MEDIUM): two more per-result unbounded prefix scans (DeliverTx + EndBlock hot paths)

- **Files**:
  - `x/audit/v1/keeper/storage_truth_divergence.go:131-153` (`storageTruthReporterDivergenceStats`) iterates the entire `ReporterStorageTruthResultPrefix(reporter)` and filters by `record.EpochID < startEpoch || > endEpoch` after JSON unmarshal — called per qualifying reporter per epoch from `ApplyReporterDivergenceAtEpochEnd` in EndBlock.
  - `x/audit/v1/keeper/storage_truth_scoring.go:553` → `storage_truth_fact_indexes.go:241-266` (`distinctNodeFailedTickets`) iterates the full `NodeStorageTruthFailurePrefix(supernode)` per failure result in DeliverTx.
- **Why it matters**: `KeepLastEpochEntries × MaxStorageProofResultsPerReport` per supernode are scanned on every failure submission. Pruning shrinks worst case but doesn't bound it.
- **Fix**: introduce `NodeStorageTruthFailureByEpochPrefix(account)+u64be(epoch)` and `ReporterStorageTruthResultByEpochPrefix(reporter)+u64be(epoch)` secondary indexes; bound scans to `[startEpoch, endEpoch+1)` (mirror the pattern at storage-proofs.go:318-350).

### NEW-A-14 (MEDIUM): Trust multiplier applied to Class B / Class C, but spec §15.4 limits it to Class A pre-recheck

- **File**: `x/audit/v1/keeper/storage_truth_scoring.go:73-81`, `539-541`
- **Evidence**:

  ```go
  bookkeeping.applyTrustScaling = isStorageTruthFailureClass(result.ResultClass) &&
      result.ResultClass != STORAGE_PROOF_RESULT_CLASS_RECHECK_CONFIRMED_FAIL &&
      result.BucketType != STORAGE_PROOF_BUCKET_TYPE_RECHECK
  ```
  `isStorageTruthFailureClass` admits HASH_MISMATCH, TIMEOUT, OBSERVER_QUORUM_FAIL, INVALID_TRANSCRIPT. After excluding RECHECK_CONFIRMED_FAIL, the multiplier still applies to TIMEOUT (Class B) and OBSERVER_QUORUM_FAIL (Class C). Spec §15.4 explicitly bounds it to Class A pre-recheck.
- **Why it matters**: a degraded reporter (R≥50) reporting a legitimate TIMEOUT or OBSERVER_QUORUM_FAIL has their suspicion contribution silently halved — the inverse of what §15.4 intends.
- **Fix**: gate `applyTrustScaling` on Class A only:

  ```go
  isClassA := result.ResultClass == HASH_MISMATCH || result.ArtifactClass == INDEX
  bookkeeping.applyTrustScaling = isClassA && result.ResultClass != RECHECK_CONFIRMED_FAIL && result.BucketType != RECHECK
  ```

### NEW-A-18 (MEDIUM): Per-result PASS reliability delta over-rewards prolific reporters (spec §15.3 is per-epoch)

- **File**: `x/audit/v1/keeper/storage_truth_scoring.go:444, 450, 456`
- **Spec ref**: §15.3 ("clean reproducible reporting epoch with no overturned fails and at least 5 results: −4")
- **Evidence**: PASS reporterReliability delta is `-4` applied per result. A reporter delivering 5 PASS results in one epoch receives −20, not the spec's −4 single-epoch reward. Worse, Class-B (TIMEOUT) gives reporterReliability=-1 (line 479), so an unreliable reporter who reports many TIMEOUTs accrues *negative* reliability score (i.e. recovery) — inverse of spec intent.
- **Fix**: zero per-result reporterReliability for PASS / TIMEOUT; emit a single −4 delta per reporter at epoch-end when `(passes_in_epoch >= 5 && no_overturned_fails)`. Mirror `ApplyReporterDivergenceAtEpochEnd` structure.

### NEW-A-13 (MEDIUM): Divergence cross-multiply uses raw uint64 — overflow risk at scale

- **File**: `x/audit/v1/keeper/storage_truth_divergence.go:65-67, 89, 92`
- **Evidence**:

  ```go
  return qualifying[i].negative*qualifying[j].total < qualifying[j].negative*qualifying[i].total
  // ...
  if entry.total == 0 || entry.negative*medianTotal <= 2*medianNeg*entry.total { continue }
  ```
  `negative` and `total` are uint64 cumulative counts. Products use uint64 wrap-around semantics on overflow, not saturation. A governance-driven `KeepLastEpochEntries` increase could silently flip the comparison and produce a wrong median ordering / wrong penalty selection.
- **Fix**: do cross-multiplies in `*big.Int` (mirror `scaleInt64TowardZero` precedent), or cap counts at a Validate-enforced ceiling.

### NEW-B-2 (MEDIUM): Healer-eligibility uses raw stored `SuspicionScore`, not decay-adjusted score (sibling-symmetry)

- **File**: `x/audit/v1/keeper/storage_truth_heal_ops.go:105-114`
- **Evidence**: This loop uses `ss.SuspicionScore` directly. Compare with enforcement.go:119 which uses `decayTowardZero(state.SuspicionScore, decay, epochDelta(epochID, state.LastUpdatedEpoch))`. Long-quiet nodes are silently barred from healing duty even though they'd be normal under decay; conversely a stale-aged probation-grade node may slip through.
- **Fix**:

  ```go
  score := decayTowardZero(ss.SuspicionScore, params.StorageTruthNodeSuspicionDecayPerEpoch,
      epochDelta(epochID, ss.LastUpdatedEpoch))
  if score >= probationThreshold { ... }
  ```

### NEW-B-8 (MEDIUM): `finalizeHealOp` does not clear failure-pattern counters after a verified heal — eligibility predicate becomes "any single new failure"

- **File**: `x/audit/v1/keeper/msg_storage_truth.go:395-404`; eligibility predicate at `storage_truth_heal_ops.go:153-159`
- **Spec ref**: §20
- **Evidence**: Verified heal touches only `DeteriorationScore`, `LastHealEpoch`, `ProbationUntilEpoch`, `ActiveHealOpId`. The predicate

  ```go
  isHealEligible := (state.DistinctHolderFailureCount >= 2) ||
      (state.LastIndexFailureEpoch > 0) ||
      (state.RecentFailureEpochCount >= 2)
  ```
  remains true forever after the first qualifying failure. Post-heal, after probation expires, any single new failure pushes D > heal-threshold and re-schedules immediately — no fresh failure-pattern accumulation required.
- **Why it matters**: Spec's "fresh start" semantic on verified heal is lost.
- **Fix**: in the verified branch, reset counters in the same write:

  ```go
  ticketState.DistinctHolderFailureCount = 0
  ticketState.RecentFailureEpochCount = 0
  ticketState.LastIndexFailureEpoch = 0
  ticketState.LastFailureEpoch = 0
  ```

### F121-F12 PARTIAL (MEDIUM): Strong-postpone band still has identical on-chain effect

- **File**: `x/audit/v1/keeper/enforcement.go:106-178, 576-588`
- **Evidence**: distinct event type (`EventTypeStorageTruthBandStrongPostpone`) is emitted, and the predicate set differs. But both bands call `setSupernodePostponed(..., postponeReasonStorageTruth)` and recovery uses identical `shouldRecoverFromStorageTruthPostponement` with one threshold + one `StorageTruthRecoveryCleanPassCount`. No behavioural distinction.
- **Fix**: distinct postpone reason for strong band; conditional recovery params (e.g. `StorageTruthStrongPostponeRecoveryCleanPassCount`).

### F121-F10 residue / F119-F3 (MEDIUM): symmetric ticket-side `ContradictionCount` still bumps on either direction without confirmation

- **File**: `x/audit/v1/keeper/storage_truth_scoring.go:418-422` (uses `storageTruthResultsContradict`)
- **Evidence**: The reporter-side bookkeeping (lines 600–626) is correctly directionalised + confirmation-guarded. But `applyTicketDeteriorationDelta` still does:

  ```go
  if state.LastResultEpoch <= epochID && state.LastTargetSupernodeAccount == result.TargetSupernodeAccount &&
      storageTruthResultsContradict(state.LastResultClass, result.ResultClass) {
      nextState.ContradictionCount = state.ContradictionCount + 1
  }
  ```
  Direction-agnostic, no confirmation guard. Per-ticket counter inflates on honest PASS-after-fail.
- **Fix**: align with the reporter-side guard — only increment on `prevFailure && currentPass && hasStorageTruthContradictionConfirmation`.

---

## LOW / informational (worth fixing but not blocking)

- **F119-F3 PARTIAL**: clean-pass-by-different-holder = −3 not implemented (PASS branch differentiates only on bucket; ignores holder identity at scoring.go:438-505).
- **NEW-A-15**: trust multiplier scales pattern-escalation bonuses (which derive from other reporters' history, not the current reporter's testimony).
- **NEW-A-17**: `epochDelta` is uint64-safe but window-checks at 256/332 are not — sibling-symmetry violation in the same file.
- **NEW-C-4 / NEW-A-19**: `pruneStorageProofTranscripts` swallows malformed records permanently with no log/metric (state corruption is unbounded).
- **NEW-B-3**: `verifierCount = 2` hardcoded in `assignStorageTruthHealParticipants:300-307`; should be a Param.
- **NEW-B-4**: insufficient-verifier path emits no event (sibling asymmetry with `EventTypeHealOpInsufficientHealers`).
- **NEW-B-5**: `linkStorageTruthRecheckTranscript` enforces single-witness recheck — confirm intent.
- **NEW-B-6 / NEW-B-9**: genesis cross-validation between audit `StorageTruthPostponements` and supernode `Postponed` state is implicit (relies on module init ordering).
- **NEW-B-7**: `GetNextHealOpID` overflow / id==0 sentinel collision unchecked.
- **NEW-A-16**: median-of-even uses lower-pair (deterministic but slight downward bias vs spec "median").
- **NF7**: `pair_rank` uses 0x00-framed concatenation — deterministic and collision-resistant, but spec §6.4 text says `H(seed||c||t||"pair")`; off-chain implementers will diverge unless spec is updated.

---

## What's verified clean ✅

- **Determinism scan** on `x/audit/v1/keeper`: 0 hits on `float|math\.Pow|time\.Now|rand\.|sort\.Float|FormatFloat`, 0 hits on `range\s+map\[` in scoring/divergence/enforcement consensus paths.
- **CRITICAL `121-F16`** float→integer conversion in `ApplyReporterDivergenceAtEpochEnd` is correctly done.
- **Geometric decay (`119-F1`)** integer-only via `factorNumerator * 1000` is correct, with proper `[1,1000]` bounds on params.
- **§20 D mutation (`120-F1`)** verified=`max(8, D/4)`, failed=`D+=15` with cooldown.
- **Empty-verifier rejection (`120-F2`)** — single-active-node networks no longer self-finalize.
- **`117-F1` / `117-F2`** R-threshold inequality and heal-threshold default 50.
- **`121-F4`** Class-A index-fail fallback is now bounded by `classAWindow`.
- **`121-F5/F9/F10`** window params consistency.
- **`121-F6`** all 7 fact-index prefixes are pruned in `prune.go::PruneOldEpochs`.
- **`121-F11`** healer/verifier exclusion of failing target + postponed nodes.
- **`121-F13`** challenger-eligibility predicate inversion fixed.
- **`122-F3`** recheck transcript hash collision check now requires bidirectional invariants.
- **`122-F4`** `KeepLastEpochEntries` lower bound + migration handler v1→v2 correctly bumps it.
- **ExportGenesis** is deterministic (all sources are prefix iterators; no map-range) — though it's missing fact indexes per NEW-C-1 above.

---

## Notes on methodology

- Detailed per-track findings stored locally at `~/work/lep6-review/round2/track-{A,B,C}-findings.md`.
- This review is a re-validation of the original 42 findings against tip `868cbc7`, not a re-review from scratch — the original line-by-line file walk happened in the first round.
- All NEW findings are spot-checked against the actual file at the cited line numbers before posting.
- This review is posted as `event=COMMENT` (not `REQUEST_CHANGES`) — the human reviewer/maintainer decides blocking.

— Zee

mateeullahmalik (commented):
# Round-2 verification by Zee — `8748065` (commit #127)

I re-checked every R2 finding against the new tip (`874806581b77b13fe26ba9fc9f5e8d6887b387bf`, commit `8748065 fix: close Zee R2 review`). Verification done by reading the actual code at the cited file:line, not by trusting the commit message.

## Result

**24/24 R2 findings: VERIFIED FIXED.** Cleared to merge from my side.

## What I confirmed (HIGH-severity items)

| Finding | Verification |
|---|---|
| **NEW-C-3** restore RegistrationFeeShareBps 2% routing | `x/action/v1/keeper/action.go:651-664` — block restored verbatim. `keeper.go:39,61,90,156` — `rewardDistributionKeeper` field/ctor/getter restored. ✅ |
| **NEW-C-1** ExportGenesis epoch-scoped state | `genesis.go:210-224` exports `RecheckEvidence`, `StorageProofTranscripts`, `NodeFailureFacts`, `ReporterResultFacts`, `FailedHealMarkers`, `EpochReports`, `ReportIndices`, `HostReportIndices`, `StorageChallengeIndices` — all 9 prefix families. ImportGenesis re-emits via existing `Set*` writers (secondary indexes rebuild naturally via `setStorageProofTranscriptRecord`). ✅ |
| **NEW-A-12** WindowStartEpoch underflow | `storage_truth_scoring.go:262, 338` use `epochDelta(epochID, state.WindowStartEpoch) >= window`. `genesis_validate.go:13-15, 26-28` reject `WindowStartEpoch > currentEpoch` for both score-state slices. ✅ |
| **NEW-B-1** EXPIRED heal cooldown | `storage_truth_heal_ops.go:43-58` — D += 15 (saturated), `ProbationUntilEpoch` advanced, `setStorageTruthFailedHeal` written, mirrors FAILED branch exactly. ✅ |

## What I confirmed (MEDIUM)

- **NEW-C-2** SSoT helper `CascadeArtifactCountsWithFallback` in `x/action/v1/types/metadata.go`, used at all 3 sites (`action.go:253`, `action_cascade.go:151,327`).
- **NEW-A-11 residue** new `NodeStorageTruthFailureEpochScanRange` and `ReporterStorageTruthResultEpochScanRange` (`keys.go:558-590`) used in both `distinctNodeFailedTickets` (fact_indexes.go:262) and `storageTruthReporterDivergenceStats` (divergence.go:221). Bounded `[startEpoch, endEpoch+1)` scans, key shape unchanged.
- **NEW-A-13** `*big.Int` cross-multiply for sort + 2x threshold (divergence.go:68-69, 90-97).
- **NEW-A-14** `applyTrustScaling` gated on `isClassA := HASH_MISMATCH || INDEX-artifact` (scoring.go:574-579). Pattern-escalation bonuses (nodeBonus/ticketBonus) added unscaled.
- **NEW-A-18** PASS/TIMEOUT per-result reporterReliability = 0; `ApplyReporterCleanEpochRecoveryAtEpochEnd` (divergence.go:149) emits single −4 once per reporter at epoch-end on >=5 PASSes with no overturned fails. Wired at abci.go:61. Uses bounded `ReporterStorageTruthResultEpochScanRange`.
- **NEW-B-2** healer eligibility uses `decayTowardZero(SuspicionScore, decay, epochDelta(epochID, ss.LastUpdatedEpoch))` (heal_ops.go:117-119) — sibling-symmetric with enforcement.go:119.
- **NEW-B-8** verified-heal branch resets `DistinctHolderFailureCount`, `RecentFailureEpochCount`, `LastIndexFailureEpoch`, `LastFailureEpoch` (msg_storage_truth.go:405-409).
- **F121-F12** distinct `postponeReasonStorageTruthStrong` (enforcement.go:18); strong marker `ap/sts/<acct>` written on entry (enforcement.go:168); recovery selects between `StorageTruthRecoveryCleanPassCount` and new `StorageTruthStrongRecoveryCleanPassCount` based on persisted reason (enforcement.go:618-619). New params have defaults (140 / 5) and `Validate()` enforces `Postpone < StrongPostpone` (params.go:737).
- **F121-F10** ticket-side `ContradictionCount` increment now guarded by `contradictionConfirmed bool` parameter passed into `applyTicketDeteriorationDelta` (scoring.go:441-446).

## LOW (also verified)

- **NEW-A-16** median-of-even now uses upper-pair (more conservative) — divergence.go:74-80, with comment.
- **NEW-B-3** `StorageTruthHealVerifierCount` param (default 2), used in heal_ops.go:323-328.
- **NEW-B-4** `EventTypeHealOpInsufficientVerifiers` emitted (heal_ops.go:238; events.go:22).
- **NEW-B-6/B-9** InitGenesis cross-validates each `StorageTruthPostponements` entry against supernode `SuperNodeStatePostponed` (genesis.go:103-115). Module init ordering confirmed: `app/app_config.go:92` shows supernode → audit → action.
- **NEW-B-7** `GetNextHealOpID` panics on malformed length and on `id==0` (storage_truth_state.go:179-188).
- **NEW-C-4** `pruneStorageProofTranscripts` now logs malformed records via `k.Logger().Error` (prune.go:253-262).
- **F119-F3 residue** cross-holder PASS bonus (−3 additional ticket-deterioration delta) implemented in scoring path with predicate on prior failure-class + different-holder.
- **NEW-A-15, A-17** auto-closed by NEW-A-14 / NEW-A-12 fixes (verified — pattern bonuses unscaled at scoring.go:83-84; both window checks now use `epochDelta`).

## Determinism scan (consensus paths)

Re-ran on `x/audit/v1/keeper/`:
- `float|math\.Pow|math\.Float|time\.Now|rand\.|sort\.Float|FormatFloat` → **0 hits**.
- `range\s+map\[` → **0 hits**.

Clean.

## Verdict

**LGTM from R2 perspective.** Every blocker, MEDIUM, and LOW from review id `4184561676` is closed at the cited file:line. No new regressions detected on the consensus paths.

Final go/no-go is the human maintainer's call — this review is `event=COMMENT` per stack-review protocol.

— Zee

mateeullahmalik (commented):
# Production-gate Round-3 review by Zee — PR #117 tip `8748065`

**Tip reviewed:** `874806581b77b13fe26ba9fc9f5e8d6887b387bf`
**Methodology:** Round-2 was a re-validation of the prior 24 findings — they're all closed at the right file:line and were verified in review id `4188647963`. **This round is a fresh deep pass for bugs introduced or unmasked by the R2-fix commit `8748065` (#127), not a re-verification of R2.** Three parallel scoped tracks (genesis I/O, scoring/heal-op, params/wiring), each combining line-by-line read + diff-walk of `868cbc7..8748065` + spec cross-check + invariant-first-coding (sibling-symmetry, write-path enumeration). Top findings spot-checked against actual file:line before posting.

## TL;DR — 3 HIGH-severity issues introduced or left LATENT

Despite the author's claim that all R2 findings are closed (verified — they are), this round finds **3 new HIGH-severity issues** plus several MEDIUMs. I do **NOT** consider this PR ready to merge yet.

| # | Severity | Status | Headline |
|---|---|---|---|
| **B-F1** | HIGH | NEW (regression in 8748065) | `ApplyReporterCleanEpochRecoveryAtEpochEnd` "no overturned fails" gate is structurally always false — fraudulent reporters whose fails are overturned still get the −4 reward |
| **C-F1** | HIGH | NEW (regression in 8748065) | `StorageTruthStrongRecoveryCleanPassCount` is **missing entirely from `Params.Validate()`** — governance can set strong-band recovery weaker than normal-band |
| **C-F3** | HIGH | LATENT (R2 missed) | `KeepLastEpochEntries` lower-bound enumeration omits `StorageTruthDivergenceWindowEpochs` (default 14) — a chain set with `KeepLast < DivergenceWindow` silently zero-stats and skips chronic-divergence penalties |

---

## CRITICAL / HIGH — must address before merge

### B-F1 (HIGH, NEW): `ApplyReporterCleanEpochRecoveryAtEpochEnd` § 15.3 "no overturned fails" gate is always false

- **File**: `x/audit/v1/keeper/storage_truth_divergence.go:195-214`; flag-writer at `storage_truth_fact_indexes.go:156-184`
- **Spec ref**: §15.3 ("clean reproducible reporting epoch with **no overturned fails** and at least 5 results: −4")
- **Evidence**:

  ```go
  // storage_truth_divergence.go:storageTruthReporterEpochPassStats
  for ; it.Valid(); it.Next() {
      var record storageTruthReporterResultRecord
      ...
      if types.StorageProofResultClass(record.ResultClass) !=
          types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_PASS {
          continue                              // ← skip non-PASS, including all FAIL records
      }
      passes++
      if record.OverturnedByRecheck {           // ← only checked on PASS records
          overturned = true
      }
  }
  ```
  `OverturnedByRecheck` is written by `markStorageTruthReporterResultRecheck` (fact_indexes.go:156-184) on the **challenged transcript's record** — i.e. a *failure-class* record, not a PASS. The PASS records this scan looks at can never have `OverturnedByRecheck=true`. So `overturned` is structurally always `false`, regardless of how many of the reporter's fails were overturned this epoch.
- **Why it matters**: A reporter who issued fraudulent FAILs that got overturned by recheck this epoch still receives the −4 §15.3 reward, as long as they also produced ≥5 PASSes. This is the exact misbehaviour §15 was designed to suppress. The trust multiplier and contradiction-overturn paths penalize the FAIL itself (+25 etc.), but the recovery gate that should cancel out the −4 reward is permanently open. NET effect: a malicious reporter accumulates net-negative reliability by parking 5+ PASSes in every epoch where they also lie.
- **Fix**: scan all classes, not just PASS, in the same iterator:

  ```go
  for ; it.Valid(); it.Next() {
      ...
      class := types.StorageProofResultClass(record.ResultClass)
      if record.OverturnedByRecheck && isStorageTruthFailureClass(class) {
          overturned = true
      }
      if class == types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_PASS {
          passes++
      }
  }
  ```

### C-F1 (HIGH, NEW): `StorageTruthStrongRecoveryCleanPassCount` is missing from `Params.Validate()`

- **File**: `x/audit/v1/types/params.go` — param introduced at lines 63, 159, 257, 319, 375, 489-491, 567 but **never validated**.
- **Spec ref**: §17 (strong-band recovery must be at least as strict as normal band)
- **Evidence**: search the entire `Validate()` body — no reference to `StorageTruthStrongRecoveryCleanPassCount`. The adjacent normal param IS validated (line 758-760 enforces `StorageTruthRecoveryCleanPassCount > 0`).
- **Why it matters**: Governance can pass `MsgUpdateParams` with `StrongRecoveryCleanPassCount = 1` and `RecoveryCleanPassCount = 10`. A node strong-postponed at score 140 would then recover faster than a node normal-postponed at 90 — **a direct inversion of LEP-6 §17 strong-band semantics**. This is consensus-affecting state-machine behaviour. The runtime fallback at `enforcement.go:619-622` substitutes 5 only when the value is zero — non-zero pathological values pass through.
- **Fix**:

  ```go
  if p.StorageTruthStrongRecoveryCleanPassCount == 0 {
      return fmt.Errorf("storage_truth_strong_recovery_clean_pass_count must be > 0")
  }
  if p.StorageTruthStrongRecoveryCleanPassCount < p.StorageTruthRecoveryCleanPassCount {
      return fmt.Errorf("storage_truth_strong_recovery_clean_pass_count must be >= storage_truth_recovery_clean_pass_count")
  }
  ```
  Add this to the existing `Validate()` block at line 758.

### C-F3 (HIGH, LATENT): `KeepLastEpochEntries` lower-bound omits `StorageTruthDivergenceWindowEpochs`

- **File**: `x/audit/v1/types/params.go:627-656` (the `requiredHistory` enumeration); consumer `x/audit/v1/keeper/storage_truth_divergence.go:42`.
- **Spec ref**: §15.3 / §22 (state retention vs lookback window)
- **Evidence**: `requiredHistory` enumerates ConsecutiveEpochsToPostpone, ActionFinalization* (3), ClassA/B/OldClassA fault windows, PatternEscalation, ContradictionWindow. **Missing: `StorageTruthDivergenceWindowEpochs`** (default 14) and `StorageTruthHealDeadlineEpochs`. The divergence scan starts at `epochID - DivergenceWindowEpochs` (divergence.go:42) — if `KeepLastEpochEntries < DivergenceWindowEpochs`, the older end of the scan range hits pruned state.
- **Why it matters**: A governance proposal setting `KeepLastEpochEntries = 7, DivergenceWindowEpochs = 14` passes `Validate()`. The divergence scan reads pruned epochs and yields silently-zero stats → no chronic-divergence penalties applied → consensus-affecting under-enforcement. Same shape as `122-F4` which the author addressed for the OldClassA window — they just missed this one.
- **Fix**: extend the enumeration:

  ```go
  if v := uint64(p.StorageTruthDivergenceWindowEpochs); v > requiredHistory {
      requiredHistory = v
  }
  if v := uint64(p.StorageTruthHealDeadlineEpochs); v > requiredHistory {
      requiredHistory = v
  }
  ```

---

## MEDIUM — should address before merge

### B-F2 (MEDIUM, LATENT): TIMEOUT_OR_NO_RESPONSE on INDEX artifact mis-classified as Class A in two places

- **File**: `x/audit/v1/keeper/storage_truth_scoring.go:273-275` (`updateNodeSuspicionHistoryFields`) and `:574-576` (trust-scaling predicate in `storageTruthBookkeepingForResult`)
- **Spec ref**: §17 (Class A = confirmed storage fault: RECHECK_CONFIRMED_FAIL, repeated HASH_MISMATCH, confirmed index artifact mismatch; Class B = TIMEOUT_OR_NO_RESPONSE)
- **Evidence**:

  ```go
  // line 273-275 (history fields)
  isClassA := result.ResultClass == HASH_MISMATCH ||
      result.ResultClass == RECHECK_CONFIRMED_FAIL ||
      result.ArtifactClass == INDEX            // ← TIMEOUT-on-INDEX matches here
  // line 282 then ALSO increments ClassBCountWindow for TIMEOUT
  // ⇒ same TIMEOUT-on-INDEX result bumps BOTH windows.

  // line 574-575 (trust-scaling predicate)
  isClassA := result.ResultClass == HASH_MISMATCH ||
      result.ArtifactClass == INDEX            // ← same shape
  ```
- **Why it matters**:
  1. **Recovery gate breakage**: `enforcement.go:635` uses `LastClassAEpoch` to forbid recovery when a Class A epoch occurred at or after the last clean pass. A genuine TIMEOUT (liveness, not storage fault) bumps `LastClassAEpoch`, blocking recovery the same as a confirmed mismatch.
  2. **Strong-postpone over-fires**: `enforcement.go:716-720` falls back to `state.ClassACountWindow >= 2`, so two TIMEOUTs on INDEX in 14 epochs trigger the strong-postpone band. Spec §17 wants 2 *Class-A faults on distinct tickets* — not liveness events.
  3. **Trust-scaling wrong band**: TIMEOUTs are reported by everyone (Class B exists because liveness reports are noisy); applying the trust multiplier here halves a degraded reporter's TIMEOUT contribution but should leave it unscaled per §15.4.
- **Fix**: INDEX artifact only escalates HASH_MISMATCH (or RECHECK_CONFIRMED_FAIL); TIMEOUT-on-INDEX stays Class B.

  ```go
  isClassA := result.ResultClass == HASH_MISMATCH ||
      result.ResultClass == RECHECK_CONFIRMED_FAIL
  // Drop the standalone "ArtifactClass == INDEX" disjunct.
  // (HASH_MISMATCH on INDEX is already +26 via the delta table — escalation is in the magnitudes, not the band.)
  ```

  Apply to both call sites.

### C-F2 (MEDIUM, NEW): `StorageTruthHealVerifierCount` has no upper bound — gov DoS surface

- **File**: `x/audit/v1/types/params.go:761-763`; consumer `x/audit/v1/keeper/storage_truth_heal_ops.go:323`
- **Evidence**: Validate enforces `> 0` only. Setting this to 10000 via gov forces every heal op to drain the eligible-supernode pool and verification-quorum compound math.
- **Fix**: add an upper cap (`<= 32` is reasonable).

  ```go
  if p.StorageTruthHealVerifierCount == 0 || p.StorageTruthHealVerifierCount > 32 {
      return fmt.Errorf("storage_truth_heal_verifier_count must be within 1..32")
  }
  ```

### C-F4 (MEDIUM, NEW): Restored 2% fee-routing has near-zero test coverage outside the new `distribute_fees_test.go`

- **File**: `testutil/keeper/action.go:368` — `ActionKeeperWithAddress` passes `nil` for `rewardDistributionKeeper`. Used by the entire pre-existing action+audit integration test suite.
- **Evidence**: `nil` short-circuits the routing block at `action.go:651`, so every existing test runs as if the 2% routing didn't exist. Only the 3 new cases in `distribute_fees_test.go` actually exercise the path.
- **Why it matters**: Behavioural divergence between `nil` and configured `RewardDistributionKeeper` is masked. A regression that breaks the routing — e.g. the bps becoming a different param, the keeper returning 0 unintentionally, the `bankKeeper.SendCoinsFromModuleToModule` failing silently — is only caught by 3 tests. Pre-existing per-supernode payout assertions silently match because routing is null.
- **Fix**: replace `nil` at line 368 with `&MockRewardDistributionKeeper{Bps: 0}` so the routing path runs (no-op when bps=0); optionally add an integration test parameterized on bps>0.

---

## LOW (documented but not blocking)

- **A-F1 (NEW)**: `GetNextHealOpID` panics on malformed/zero counter; sibling `GetNextEvidenceID` recovers gracefully. Sibling-asymmetry. Storage_truth_state.go:182-191.
- **A-F2 (NEW)**: Two ad-hoc `endEpoch+1` range builders in `storage_truth_fact_indexes.go:299-301, 333-336` lack the `^uint64(0)` wrap-guard that the named `*EpochScanRange` helpers have.
- **A-F3 (NEW)**: ExportGenesis returns error on `NextHealOpId == 0` — dead branch given A-F1's panic.
- **A-F4 (LATENT)**: `ValidateScoreStatesGenesis` is missing the new `TicketArtifactCountState` family — sibling-asymmetry with the other 3 score-state families.
- **A-F5 (NEW)**: `importStorageProofTranscriptForGenesis` silently drops unknown JSON fields — forwards-compat hazard.
- **A-F6 (NEW)**: New round-trip test in `genesis_test.go` never seeds `NodeFailureFacts` / `ReporterResultFacts`; the highest-risk parsers are unmodelled in tests.
- **B-F3 (LATENT)**: `hasIndependentReporterPassInWindow` / `hasCleanRecheckInWindow` overflow when `endEpoch == MaxUint64` (same root as A-F2).
- **B-F4 (LATENT)**: `setStorageTruthFailedHeal` silent no-op when `HealerSupernodeAccount == ""` masks scheduler invariant violation.
- **B-F5 (NEW)**: `ApplyReporterCleanEpochRecoveryAtEpochEnd` skips reporters with no `ReporterReliabilityState` row — fresh-reporter case undercounts in dashboards.
- **B-F6 (NEW)**: Cross-holder PASS bonus tests parameterise prior class as PASS only — TIMEOUT/OBSERVER_QUORUM_FAIL/INVALID_TRANSCRIPT branches untested.
- **C-F5 (LATENT)**: `Validate()` allows `Postpone == StrongPostpone` (`<=`), collapsing the strong band into a no-op.
- **C-F6 (NEW)**: `CascadeArtifactCountsWithFallback` returns `(0,0)` if both counts AND `RqIdsIds` are empty; protected by `Process` validation today, but defense-in-depth could panic earlier.
- **C-F7 (LATENT)**: Module ordering (supernode → audit → action) is correct in `app/app_config.go` but unpinned by tests; future refactor risk.
- **C-F9 (NEW)**: No v2→v3 migration. Default-on works for the new params via `WithDefaults()`, but combined with C-F1, governance can intentionally set `StrongRecoveryCleanPassCount=0` and pass `Validate()` — only `WithDefaults()` rewrites it inside `SetParams`. Fragile interaction.

---

## What's verified clean ✅

- All 24 R2 findings remain FIXED (re-spot-checked).
- Determinism scan on `x/audit/v1/keeper`: **0 hits** for `float|math.Pow|math.Float|time.Now|rand.|sort.Float|FormatFloat`; 0 hits for `range\s+map\[`.
- Genesis parser ⇄ `*Key()` encoder symmetry verified for all 6 new families (slash positions, 0x00 separators, u64be epoch).
- Iterator key/value safety: all parsers convert to `string(...)` before `it.Next()`.
- ImportGenesis ordering: score-state imports (lines 68–95) → postponement reconciliation (103–115) → fact-index imports (118–148). Fact-index re-emitters write only KV records; no score-state mutation. ✓
- Strong-postpone marker IS cleared on recovery via `clearStorageTruthPostponedAtEpochID` (postponement_state.go:28-32).
- Big.Int `Cmp` cross-multiply ordering is deterministic (math/big spec).
- Bounded-scan inclusivity at all in-scope callers is correct.
- ABCI ordering: `ApplyReporterDivergenceAtEpochEnd` runs BEFORE `ApplyReporterCleanEpochRecoveryAtEpochEnd`. The −4 recovery is applied to the post-divergence score; no double-decay (LastUpdatedEpoch already advanced).
- Migration handler is unchanged but safe — new params land via blob-zero-decode + `WithDefaults()`, defaults are mutually consistent.

---

## Recommendation

**Not ready to merge.** Address the 3 HIGH findings (B-F1, C-F1, C-F3), the 3 MEDIUMs (B-F2, C-F2, C-F4), and consider the LOWs case-by-case. The HIGHs are all small one-to-five-line fixes with corresponding tests. None requires architectural change. After the fix:

1. Add a per-finding regression test (especially for B-F1 — assert that a reporter with 5 PASS + 1 overturned-by-recheck FAIL does NOT receive −4).
2. Re-run `make verify` and the audit/action integration suite.
3. Ping me to re-validate.

Posted as `event=COMMENT`. Final go/no-go is the human maintainer's call.

— Zee

mateeullahmalik (commented):
# Round-3 verification by Zee — PR #117 tip `ef5991b` (commit #128)

I re-checked all 20 R3 findings against the new tip (`ef5991ba525ab7b498a9ff4b22d45e0aa9815a90`, commit `ef5991b fix(lep6): close Zee R3 foundation review findings`). Verification by reading actual code at the cited file:line, not the commit message.

## Result

**20/20 R3 findings VERIFIED FIXED** at the right file:line, with regression tests included.

## HIGH (3/3 verified)

| Finding | Verified at |
|---|---|
| **B-F1** §15.3 overturned-fail gate | `storage_truth_divergence.go:240-268` — single iterator over all classes; `OverturnedByRecheck && isStorageTruthFailureClass(class)` flips overturn; PASS class counted separately. New regression test `storage_truth_overturn_gate_test.go` (106 lines) asserts 5 PASS + 1 overturned HASH_MISMATCH does NOT receive −4. ✅ |
| **C-F1** `StrongRecoveryCleanPassCount` Validate gap | `params.go:781-787` — `> 0` check + `>= RecoveryCleanPassCount` ordering invariant. New `params_r3_validate_test.go` (95 lines) covers both. ✅ |
| **C-F3** `KeepLastEpochEntries` lookback | `params.go::Validate()` `requiredHistory` enumeration extended with `StorageTruthDivergenceWindowEpochs` and `StorageTruthHealDeadlineEpochs`. Migration handler `module/migrations.go::NewMigrateV1ToV2` was extended in-place (v2 hasn't shipped to mainnet) to bump `KeepLastEpochEntries` to cover both windows; new `migrations_test.go`. ✅ |

## MEDIUM (3/3 verified)

| Finding | Verified at |
|---|---|
| **B-F2** TIMEOUT-on-INDEX Class-A misclassification | Both call sites fixed: `storage_truth_scoring.go:277-278` (history-fields) drops `ArtifactClass==INDEX` disjunct, isClassA only on `HASH_MISMATCH \|\| RECHECK_CONFIRMED_FAIL`. `:578` (trust-scaling predicate) similarly narrowed. Comment cites CP-R3 B-F2 explicitly. ✅ |
| **C-F2** `HealVerifierCount` upper bound | `params.go:791-793` — `> 0 && <= 32`. ✅ |
| **C-F4** testutil rewardDist mock | `testutil/keeper/action.go:368` — `&MockRewardDistributionKeeper{Bps: 0}` replaces the `nil` short-circuit, so all pre-existing action+audit tests now exercise the routing branch as no-op. ✅ |

## LOW (14/14 verified)

- **A-F1** `GetNextHealOpID` graceful recovery (`storage_truth_state.go:172-208`) — three-way fallback (nil / wrong-len / id==0 → `deriveNextHealOpID` scans `st/ho/` for max). Sibling-symmetric with `GetNextEvidenceID`. ✅
- **A-F2 / B-F3** ad-hoc range builders → named ScanRange helpers (`keys.go:672-722`) `ReporterStorageTruthResultByTargetEpochScanRange` and `TranscriptByTargetBucketEpochScanRange`, both `^uint64(0)`-safe via `prefixEnd(base)` fallback. ✅
- **A-F3** `ExportGenesis` `NextHealOpId == 0` branch is now reachable (consistent with the graceful A-F1 fix). ✅
- **A-F4** `ValidateScoreStatesGenesis` extended with `WindowStartEpoch > currentEpoch` checks AND a new `TicketArtifactCountStates` block (`genesis_validate.go:36-44`) rejecting empty ticket-id and `(0,0)` counts. ✅
- **A-F5** `importStorageProofTranscriptForGenesis` uses `json.Decoder.DisallowUnknownFields()` (`storage_truth_fact_indexes.go:404-410`). ✅
- **B-F4** `setStorageTruthFailedHeal` returns explicit error on empty supernode/ticket; both call sites (`storage_truth_heal_ops.go:56-58` and `msg_storage_truth.go:418-420`) propagate. ✅ *Subtle note below.*
- **B-F5** `ApplyReporterCleanEpochRecoveryAtEpochEnd` now unions `GetAllReporterReliabilityStates` with the new `storageTruthReporterAccountsForEpoch` helper (`divergence.go:163-178`), then sorts the union for determinism — fresh reporters are no longer skipped. ✅
- **B-F6** `storage_truth_cross_holder_pass_test.go:113-115` parameterises prior-class over TIMEOUT / OBSERVER_QUORUM_FAIL / INVALID_TRANSCRIPT. ✅
- **C-F5** `Postpone < StrongPostpone` strict (`params.go:747-754`, `>=` rejected). ✅
- **C-F6** `CascadeArtifactCountsWithFallbackStrict` returns explicit error when both counts and `RqIdsIds` are zero (`x/action/v1/types/metadata.go:22-39`); all 3 consensus-path callers in `action.go:253` and `action_cascade.go:150,329` use the strict variant. Non-strict variant preserved for backward compat. ✅
- **C-F7** new pin test `app/lep6_module_order_test.go` asserts supernode→audit→action across `genesisModuleOrder`, `beginBlockers`, `endBlockers`. ✅
- **C-F9** Migration handler extended in-place (`module/migrations.go`) since v2 has not shipped to mainnet — `KeepLastEpochEntries = max(OldClassA, Divergence, HealDeadline)`. ✅

## Determinism scan (consensus paths)

`x/audit/v1/keeper/`: 0 hits for `float|math.Pow|math.Float|time.Now|rand.|sort.Float|FormatFloat`; 0 hits for `range\s+map\[`. The two new map iterations introduced by R3 fixes (`divergence.go:174` and `:231` over `reporterSet`) are both followed by `sort.Strings(reporters)` — deterministic. ✅

## One subtle note (B-F4 strict-error propagation)

The B-F4 fix changes `setStorageTruthFailedHeal` from a silent no-op to a hard error on empty `supernodeAccount`. Both call sites (`expireStorageTruthHealOpsAtEpochEnd` and `finalizeHealOp`) now propagate the error, which means the EndBlock returns an error if any non-final HealOp has empty `HealerSupernodeAccount`. The runtime scheduler at `storage_truth_heal_ops.go:247-263` guarantees non-empty healer (it `continue`s when assignment fails before constructing the `HealOp`), so this is a safe state machine invariant. **However** `ImportGenesis` at `genesis.go:88` re-emits `genState.HealOps` via `SetHealOp` without validating `HealerSupernodeAccount != ""`, so a malformed genesis with an empty-healer non-final HealOp would chain-halt on the first epoch end. This is governance-controlled (genesis only, no runtime path can produce the malformed state), but worth a defense-in-depth tweak: add a `HealerSupernodeAccount != ""` check inside the heal-op import loop in `genesis.go:88-95` for non-final-status HealOps. Not blocking — flagging for a follow-up.

## Tests run

The commit message claims:
- `go build ./...` ✅ (assumed; not re-run here)
- `go test ./x/audit/v1/{types,keeper,module} ./x/action/v1/{types,keeper} ./app` ✅
- `go test ./x/...` ✅
- `go test ./tests/integration/{action,audit}` ✅
- `go test -tags=system ./tests/system/...` ✅
- `go test -tags=system_test -timeout=1800s -v . from tests/systemtests` — passed in 1031s ✅

## Verdict

**LGTM from R3 perspective. Cleared for merge from my side.** All 20 findings closed at the right file:line with corresponding regression tests, no determinism regressions, no new HIGH/MEDIUM issues introduced. The single LOW follow-up (genesis-import healer-account validation) is a defense-in-depth improvement, not a blocker.

Final go/no-go is the human maintainer's call.

— Zee

mateeullahmalik (commented):
# FINAL Production-Gate Review by Zee — PR #117 tip `ef5991b`

**Tip reviewed:** `ef5991ba525ab7b498a9ff4b22d45e0aa9815a90`
**Mandate:** Final pre-merge gate. User wants explicit assurance no consensus problem / chain halt / determinism issue.

## Methodology

Three parallel tracks, each focused exclusively on consensus-safety:
- **Track A**: Determinism (float, time, rand, map iteration), chain-halt panic class, KV-iteration ordering, divisor/modulus on data, slice-bound safety. Full grep-based audit + line-by-line read of every consensus-path file in `x/audit/v1/{keeper,types}`.
- **Track B**: Genesis InitGenesis/ExportGenesis round-trip determinism, migration safety, `Params.Validate()` completeness, `MsgUpdateParams` malleability surface.
- **Track C**: Cross-module integration (action↔audit↔supernode), ABCI ordering, DistributeFees money flow, FinalizeAction→audit hook safety, IBC surface, OOG bounds.

R3 closure (review id `4191762658`) was already verified in commit `ef5991b`. This round did not re-validate R3 — it hunted what we and the author missed.

---

## Verdict

# 🟡 SAFE TO MERGE WITH PRE-MAINNET FOLLOW-UPS

**No CRITICAL / chain-halt / in-block determinism issues found.** The PR is consensus-safe for live chain operation.

However, **5 HIGH issues** are open that you should track before any production-restart workflow or before significant cascade volume materializes. None blocks merge of the PR; all are safe to fix in follow-up PRs as long as the team is aware before mainnet.

---

## What's verified clean ✅ (the consensus-critical surface)

| Class | Result |
|---|---|
| **Float / Pow / Sqrt / FormatFloat / sort.Float** | 0 hits in consensus paths. The 5 `float64` hits in `enforcement.go::violatesMinFree/compliesMinFree` and `state.go:52` operate on `HostReport.{Cpu,Mem,Disk}UsagePercent` proto fields decoded via `math.Float64frombits` (bit-exact), used only for `<`, `==`, `<=` and one `100.0 - x` subtraction. Deterministic per IEEE-754 + Go spec. No `sort.Float`, no float-keyed map, no `FormatFloat`. ✅ |
| **time.Now / rand / crypto/rand** | 0 hits in scope. ✅ |
| **Map iteration (`for x := range map[]`)** | All 5 hits verified order-safe: either sorted before consumption (`storage_truth_divergence.go:174,231` → `sort.Strings`), used as a min-by-deterministic-rank with lex tiebreak (`audit_peer_assignment.go:156`), or consumed only as membership test in error-only paths (`msg_submit_epoch_report_storage_proofs.go:359,389`). ✅ |
| **panic / Must… in consensus paths** | Only `cdc.MustUnmarshal` against bytes just written via `cdc.Marshal` — round-trip safe by construction. No `MustNewDecFromStr`, no raw `panic(`, no `MustParse`. ✅ |
| **`binary.BigEndian.Uint64(bz)` length checks** | All 13 call sites verified with explicit `len(bz) != 8` or `len(key) >= len(prefix)+8` guards. ✅ |
| **Division/modulo with data-driven divisor** | All 11 call sites have either explicit `if x == 0 return` guards or are reached only after `Params.Validate()` enforces `> 0` on the param. `decayTowardZero` denominator is hardcoded 1000; factor bounded `[1,1000]` by Validate. `scaleInt64TowardZero` uses `*big.Int` to avoid overflow + early-returns on `denominator <= 0`. ✅ |
| **Slice indexing from data** | Every `body[split+1:]`, `key[len(prefix)+8:]`, `obs.PortStates[i]` is preceded by an explicit length check. ✅ |
| **EndBlock ordering** | `app/app_config.go:156-176` gives supernode → audit → action across `genesisModuleOrder`, `beginBlockers`, `endBlockers`. Pinned by new `app/lep6_module_order_test.go`. Audit reads supernode active set after supernode EndBlock has settled. Action's `DistributeFees` runs only on tx path, not EndBlock. No circular hook dependency. ✅ |
| **DistributeFees money flow** | All `Sub` operations are bounded by their preceding `Mul` (bps ≤ 10000 enforced). No underflow. The ordering reward → foundation → per-SN is sequential against residual fee, not original — composition stays consistent even when shares conceptually exceed 100%. ✅ |
| **`SetStorageTruthTicketArtifactCounts`** idempotent re-write returns nil; mismatch returns error. Caller `FinalizeAction` reverts atomically on error. No half-written state. ✅ |
| **`storageTruthAssignmentHash`** (audit_peer_assignment.go:232) | Single SHA-256 over `seed || (0x00 || part_i)*`. Bech32 part strings cannot contain 0x00 → unambiguous framing. Seed is fixed-length so prefix-length is implicit. No collision surface. ✅ |
| **Eligible challenger / healer / verifier pools** | All read from `anchor.ActiveSupernodeAccounts` (frozen at epoch start, sorted at write time) or `sort.Strings(supernodeKeeper.GetAllSuperNodes(...))`. Selection comparators use `bytes.Compare` on SHA-256 outputs with lex-tiebreak on bech32 strings (ASCII = byte compare). Empty pool returns empty slice, no panic. ✅ |
| **OOG per-tx** | `MsgSubmitEpochReport.StorageProofResults` capped at 16 (`MaxStorageProofResultsPerReport` validated in handler). Per-result inner loops are O(results × constant). Recheck-evidence handles one record per Msg. ✅ |
| **IBC surface** | No new packet types, ack handlers, or channel openings. No `proto/ibc/*` changes. New audit state lives under audit module's own KV prefixes; no IBC commitment-path implications. ✅ |
| **R3 fixes (20 of them)** | All confirmed still in place at this tip. Determinism scan still 0 hits. ✅ |

---

## HIGH severity — must address before mainnet (NOT blocking this merge)

### F-B1 (HIGH, MIGRATION/DETERMINISM): Strong-postpone band marker `ap/sts/` is silently dropped on `lumerad export → init`

- **File:Line**: `proto/lumera/audit/v1/genesis.proto::StorageTruthPostponement` has only `{supernode_account, postponed_at_epoch_id}` — no `strong` flag. `x/audit/v1/keeper/genesis.go:103-115, 211` round-trips only those two fields.
- **Reproduction**: chain runs, supernode crosses `StorageTruthNodeSuspicionThresholdStrongPostpone` (140), enforcement.go:168 writes `ap/sts/<acct>=1`. Operator runs `lumerad export`. New chain bootstraps from this genesis. `hasStorageTruthStrongPostponeMarker` returns false → recovery now requires `RecoveryCleanPassCount` (default 3) instead of `StrongRecoveryCleanPassCount` (default 5). Strong-postponed node recovers ~40% faster than spec.
- **Why not consensus-halt today**: every validator restarting from the same exported genesis loses the same data, so the resulting chain is internally consistent — just semantically wrong vs. source chain. Live in-place upgrades (KV intact) are unaffected.
- **Fix**: add `bool strong_postpone` to `StorageTruthPostponement` proto; export by reading `ap/sts/` companion key; import via `setStorageTruthStrongPostponeMarker`.

### F-B2 (HIGH, MIGRATION/DETERMINISM): Action-finalization postponement marker `ap/af/` not in genesis at all

- **File:Line**: `x/audit/v1/keeper/action_finalization_postponement_state.go:10-30` — set/get/clear. No `GetAllActionFinalizationPostponements` exists; no field on `GenesisState`; no import loop in `genesis.go`. Consumer `enforcement.go:251` (recovery routing).
- **Reproduction**: AF-postponed supernode → export → import → marker absent → supernode evaluated for recovery via host/peer-port path, bypassing `ActionFinalizationRecoveryEpochs` and `ActionFinalizationRecoveryMaxTotalBadEvidences`. Recovery rules silently change.
- **Fix**: add `repeated ActionFinalizationPostponement` to `GenesisState`; mirror the storage-truth pattern.

### F-B3 (HIGH, MIGRATION/DETERMINISM): Evidence per-epoch counters `eve/` not exported/imported

- **File:Line**: `x/audit/v1/keeper/evidence_epoch_count.go:10-26`. Writer `incrementEvidenceEpochCount` is called only from `evidence.go:87` (live). Consumers `enforcement.go:356, 388-389` drive AF postponement and recovery. **Zero** Genesis I/O for the `eve/` prefix.
- **Reproduction**: existing chain has accumulated `eve/<epoch>/<subject>/<type>` counters used by `shouldPostponeForActionFinalizationEvidence`. Export → import → all counters drop to 0. Postponement gating becomes a no-op until counters re-accumulate; suspect supernodes recover prematurely.
- **Fix**: either (a) reconstruct counters by walking imported `Evidence` slice in InitGenesis, or (b) round-trip via explicit `repeated GenesisEvidenceEpochCount` field.

### F-B4 (HIGH, MIGRATION/DETERMINISM): HealOp verifications `st/hov/` not exported/imported

- **File:Line**: `x/audit/v1/keeper/storage_truth_state.go:265-305` (set/get/getAll). `genesis.HealOps` round-trips the heal-op record itself but the verification sub-keys are dropped.
- **Reproduction**: HealOps in IN_PROGRESS / VERIFYING with partial verifications. Export → import → all per-verifier votes vanish; quorum count restarts at 0. Heal-ops one verification away from VERIFIED reset to needing the full `StorageTruthHealVerifierCount` from scratch.
- **Fix**: add `repeated GenesisHealOpVerification {uint64 heal_op_id; string verifier; bool verified;}` to GenesisState; export via `GetAllHealOpVerifications` per non-final heal-op.

### F-C1 (HIGH, OOG): `TicketDeteriorationState` is iterated every epoch-end but never pruned — slow-burn EndBlock OOG halt risk

- **File:Line**: `x/audit/v1/keeper/storage_truth_heal_ops.go:105` — `ticketStates, _ := k.GetAllTicketDeteriorationStates(ctx)` runs unconditionally at every epoch-end. `grep -rn "TicketDeteriorationState" x/audit/v1/keeper/ | grep Delete` returns **zero** delete sites. `prune.go::PruneOldEpochs` only handles epoch-leading prefixes; `st/td/` is keyed by `ticket_id`, not epoch. `TicketArtifactCountState` (`st/tac/`) has the same shape.
- **Reproduction**: every cascade ticket experiencing a storage-proof event creates a `TicketDeteriorationState` row that is never deleted. Over chain lifetime, every epoch-end EndBlocker iterates and proto-unmarshals N rows, sorts, filters. An attacker can amplify by spamming cheap cascade actions (or naturally accumulates with mainnet usage). When per-block work exceeds consensus timeout or hits OOM on smaller validators, validators fork on liveness.
- **Why not caught by Validate / ante**: there is no per-epoch cap on iteration. `StorageTruthMaxSelfHealOpsPerEpoch` only caps how many heal-ops are scheduled, not how many ticket states are scanned. The full prefix scan runs unconditionally before the cap is applied.
- **Why not consensus-halt today**: cascade volume on devnet/testnet is low. Becomes a real concern at mainnet scale or under deliberate amplification.
- **Fix**: prune `st/td/` rows where `LastUpdatedEpoch + KeepLastEpochEntries < currentEpoch && DeteriorationScore == 0 && ActiveHealOpId == 0` from `PruneOldEpochs`. Same treatment for `st/tac/` once finalize-state is settled. Alternative: maintain a `(score_bucket || ticket_id)` secondary index and only iterate the high-score band.

---

## MEDIUM severity (track but not blocking)

### F-B5 (MEDIUM, MIGRATION/MALLEABILITY): `NewMigrateV1ToV2` writes params back without re-running `Validate()`

- **File:Line**: `x/audit/v1/module/migrations.go:14-41`; `keeper.SetParams` does NOT call `Validate`.
- **Reproduction**: a v1 chain with `Postpone=200, StrongPostpone=0` → migration runs `WithDefaults` (StrongPostpone=140) → `SetParams` writes the now-invalid `Postpone(200) >= StrongPostpone(140)` combo. Runtime never re-validates, the strong band collapses.
- **Fix**: call `params.Validate()` at the end of the migration handler (and ideally inside `keeper.SetParams`).

---

## LOW severity (informational / defense-in-depth)

| # | Class | Headline |
|---|---|---|
| F-A1 | DETERMINISM (theoretical) | `decayTowardZero` int64 multiplication wraps if score ever reached `MaxInt64` (deltas ≤ 26 per result + saturation makes this unreachable in practice) |
| F-A2 | DETERMINISM (informational) | `float64` HostReport usage fields — pure comparisons, bit-exact, deterministic, but flagged for policy. Could be migrated to `uint32` |
| F-B6 | MALLEABILITY | `MaxProbeTargetsPerEpoch == 0` not rejected by Validate (relies on WithDefaults); JSON genesis with explicit zero would silently disable probe coverage |
| F-B7 | MALLEABILITY | `KeepLastEpochEntries` and window params have no upper bound; `MaxUint64` accepted — pruning becomes a no-op forever |
| F-B8 | MALLEABILITY | `ScChallengersPerEpoch` not bounded; sentinel-zero overload undocumented |
| F-B9 | INFO | `ConsensusVersion=2` + `RegisterMigration(1, ...)` + `app/upgrades/v1_11_1` audit pinning verified consistent. v1.12.0 doesn't touch audit |
| F-B10 | MIGRATION | `importStorageProofTranscriptForGenesis` uses `DisallowUnknownFields` — strict but means cross-version genesis imports break. Other fact-index imports use raw-bytes pattern; recommend mirroring |
| F-B11/12 | INFO | No duplicate-ID assertions for `EvidenceId` / `HealOpId` in genesis validator |
| F-C2 | MALLEABILITY | `RegistrationFeeShareBps + FoundationFeeShare > 100%` not validated cross-module — sequential subtraction stays consistent, no underflow, cosmetic only |
| F-C3 | MALLEABILITY | `validateDecString` accepts negative LegacyDec → governance can set `(superNode=2.0, foundation=-1.0)` → `sdk.NewCoin` panics at FinalizeAction → per-tx revert (NOT halt). Add non-negative guard |

---

## Pre-mainnet checklist

1. **Address F-C1 first** (state-bloat OOG) — schedule pruning for `st/td/` and `st/tac/` BEFORE mainnet sees significant cascade volume. This is the only HIGH that can manifest on a live chain over time.
2. **Address F-B1 through F-B4 before any export-import workflow is exercised in production** — testnet hand-offs, disaster-recovery genesis-export, fork-from-state operations all hit these paths. They're mechanically straightforward (proto field add + GetAll/Set helper + InitGenesis loop).
3. **Add `Validate()` to `keeper.SetParams`** (closes F-B5).
4. **Tighten Validate** for F-B6/B7/B8/C3 — cheap defensive fixes against future gov mistakes.

---

## Summary

| Track | CRITICAL | HIGH | MEDIUM | LOW |
|---|---|---|---|---|
| A — Determinism / chain-halt / KV iteration | 0 | 0 | 0 | 2 |
| B — Genesis / migration / param malleability | 0 | 4 | 1 | 7 |
| C — Cross-module / EndBlock / DistributeFees | 0 | 1 | 0 | 2 |
| **Total** | **0** | **5** | **1** | **11** |

**No consensus-affecting issues at runtime. No deterministic chain-halt at runtime. No determinism issues at runtime.** The 5 HIGH findings are all latent issues that materialize either (a) on `lumerad export | init-from-genesis` workflows (F-B1..B4) or (b) under sustained cascade volume / chain age (F-C1). Address them before mainnet but they do not block this merge.

Posted as `event=COMMENT`. Final go/no-go is the human maintainer's call.

— Zee

mateeullahmalik (approved):

PR #121, opened 2026-04-21
audit: implement storage truth enforcement and self-healing lifecycle

j-rafique:


x/audit/v1/keeper/storage_truth_scoring.go
```diff
@@ -167,25 +187,88 @@ func (k Keeper) applyNodeSuspicionDelta(
...
+
+	isFailure := isStorageTruthFailureClass(result.ResultClass)
+	isPass := result.ResultClass == types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_PASS
+
+	if isPass {
+		state.CleanPassCount++
+		state.LastCleanPassEpoch = epochID
+	}
+
+	if isFailure {
```
ghost, line 226: `CleanPassCount` is incremented on every PASS but never reset when a failure occurs. The recovery gate in `shouldRecoverFromStorageTruthPostponement` checks `state.CleanPassCount >= requiredPasses` (default 3). Because the counter accumulates over the node's entire lifetime, a node that earned 3+ clean passes before any violations would immediately satisfy the recovery gate once decay brings its score below the watch threshold -- even right after being postponed for severe storage-truth faults. The spec's intent ("recovery requires N clean passes") likely means N clean passes since the last failure or since postponement, not N total historical passes. Consider resetting `CleanPassCount` to 0 when a failure is recorded.

```suggestion
	if isPass {
		state.CleanPassCount++
		state.LastCleanPassEpoch = epochID
	}

	if isFailure {
		// Reset clean pass streak -- recovery requires consecutive clean passes after failures.
		state.CleanPassCount = 0
```

<sub>[Fix it with Roo Code](https://app.roocode.com/fix-issue/?repo=LumeraProtocol/lumera&prNumber=121&fixId=708f5d38eb) or mention @roomote and request a fix.</sub>

x/audit/v1/keeper/storage_truth_scoring.go
```diff
@@ -167,25 +187,88 @@ func (k Keeper) applyNodeSuspicionDelta(
...
+		switch result.ResultClass {
+		case types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_HASH_MISMATCH,
+			types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_RECHECK_CONFIRMED_FAIL:
+			state.ClassACountWindow++
+			state.LastClassAEpoch = epochID
+		case types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_TIMEOUT_OR_NO_RESPONSE:
+			state.ClassBCountWindow++
+			state.LastClassBEpoch = epochID
+		}
+	}
```
ghost, line 261: `ClassACountWindow` only increments for `HASH_MISMATCH` and `RECHECK_CONFIRMED_FAIL`, but the Class A definition in `storageTruthIsClassAFault` (used by enforcement predicates) also includes any failure with `ARTIFACT_CLASS_INDEX` regardless of result class. When fact indexes are empty, the enforcement fallback path in `storageTruthPostponePredicatesMet` uses `state.ClassACountWindow` to evaluate postpone predicates. This means INDEX artifact failures (e.g. `TIMEOUT_OR_NO_RESPONSE` on an INDEX artifact) would not be counted as Class A in the fallback, potentially preventing a justified postponement.

```suggestion
		case types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_HASH_MISMATCH,
			types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_RECHECK_CONFIRMED_FAIL:
			state.ClassACountWindow++
			state.LastClassAEpoch = epochID
		case types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_TIMEOUT_OR_NO_RESPONSE:
			state.ClassBCountWindow++
			state.LastClassBEpoch = epochID
		default:
			// Other failure result classes on INDEX artifacts are also Class A.
		}
		if result.ArtifactClass == types.StorageProofArtifactClass_STORAGE_PROOF_ARTIFACT_CLASS_INDEX {
			if result.ResultClass != types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_HASH_MISMATCH &&
				result.ResultClass != types.StorageProofResultClass_STORAGE_PROOF_RESULT_CLASS_RECHECK_CONFIRMED_FAIL {
				state.ClassACountWindow++
				state.LastClassAEpoch = epochID
			}
		}
```

<sub>[Fix it with Roo Code](https://app.roocode.com/fix-issue/?repo=LumeraProtocol/lumera&prNumber=121&fixId=199a3a6aff) or mention @roomote and request a fix.</sub>

x/audit/v1/keeper/storage_truth_scoring.go
```diff
@@ -475,21 +714,61 @@ func boolToUint64(v bool) uint64 {
...
 	}
-	if decayTotal >= -score {
+	if result > 0 {
 		return 0
 	}
-	return score + decayTotal
+	return result
 }
 
 func mulInt64ByUint64Saturated(v int64, m uint64) int64 {
```
ghost, line 774: `mulInt64ByUint64Saturated` is defined but has no callers in the codebase. It appears to be leftover from a previous iteration of the scoring logic. Consider removing it to reduce dead code.

<sub>[Fix it with Roo Code](https://app.roocode.com/fix-issue/?repo=LumeraProtocol/lumera&prNumber=121&fixId=e4f8dda333) or mention @roomote and request a fix.</sub>

mateeullahmalik:
# Production-gate review by Zee — 15 findings

**Methodology**: full file-by-file read of every non-generated changed file in this PR's diff (pr-121 vs its base branch), cross-checked against:
- LEP-6 spec (Notion source of truth)
- `invariant-first-coding` skill (write-path enumeration, sibling symmetry, single source of truth, post-fix re-audit)
- Cosmos SDK consensus discipline (no float, no map iteration, bounded EndBlock, genesis round-trip, errorsmod wrapping)

**Status legend**: each finding's status is computed at the **PR #122 stack-tip** (consensus-gap-fixes commit `a51c439`), so 'FIXED' means a downstream PR in the stack already addresses it; 'OPEN' means it is still present at the tip and must be fixed before merge / Phase-2 activation. Severity rubric in the charter (`~/work/lep6-review/ctx/charter.md`): CRITICAL = consensus halt / state corruption / non-determinism in ABCI; HIGH = spec mismatch with economic impact, missing genesis round-trip, replay enabler; MEDIUM = invariant asymmetry without immediate exploit, unbounded loop with practical bound, missing param validation.

**Severity breakdown:** CRITICAL=1, HIGH=7, MEDIUM=7

---

## 121-F16 — DETERMINISM: float64 arithmetic in EndBlock consensus path drives state mutation
- **Severity**: `CRITICAL`
- **File**: `x/audit/v1/keeper/storage_truth_divergence.go`
- **Lines**: 32 (struct field), 50 (`negRate := float64(stats.negative)/float64(stats.total)`), 65-69 (`sort.Float64s`, `medianFloat64`), 73 (`entry.negRate <= 2.0*medianNegRate`), 97-98 (`strconv.FormatFloat` in event attribute)
- **Status at PR #122 tip**: OPEN at PR #122 head — must fix before mainnet activation. Float in Cosmos consensus path is an anti-pattern even when IEEE-754 is nominally deterministic; Go does not contractually guarantee bit-exact float ops across compiler versions/optimizer flags/CPU instruction selection (e.g. FMA fusion). Event attributes built with `strconv.FormatFloat` are written into the consensus event log and any cross-version difference becomes an apphash mismatch.
- **What**: `ApplyReporterDivergenceAtEpochEnd` runs every EndBlock at epoch end. It computes per-reporter `negRate := float64(neg)/float64(total)`, sorts a `[]float64` of rates, takes the float median, and the predicate `entry.negRate <= 2.0*medianNegRate` gates a +8 reliability penalty (state mutation via `applyReporterReliabilityDelta`). It additionally emits `reporter_neg_rate` and `median_neg_rate` event attributes via `strconv.FormatFloat(..., 'f', 4, 64)`, which become part of the consensus event log.
- **Suggested fix**: Replace with integer / `cosmossdk.io/math.LegacyDec` arithmetic. The spec predicate `neg_rate_reporter > 2.0 * neg_rate_network` can be expressed exactly as cross-multiplied integers: `negR * totN > 2 * negN * totR` (after sorting reporters by `(negR, totR)` for the median). Replace `medianFloat64(rates)` with a deterministic median over `(numerator, denominator)` integer pairs. Replace event attribute formatting with integer/string forms (e.g. emit `negative` and `total` separately, let off-chain consumers compute the rate). This is a hard determinism gate — flag must block merge/activation.

## 121-F1 — Recheck PASS double-counts original-reporter penalty: +12 (contradiction) + +25 (explicit) = +37 vs spec +25
- **Severity**: `HIGH`
- **File**: `x/audit/v1/keeper/msg_storage_truth.go`
- **Lines**: 117-141 (synthesis call) interacting with storage_truth_scoring.go:575-587
- **Status at PR #122 tip**: OPEN — please confirm at PR #122 tip. Either bypass `storageTruthBookkeepingForResult` for `BucketType==RECHECK`, or apply only one of the deltas in the recheck handler.
- **What**: `SubmitStorageRecheckEvidence` synthesises a PASS result, applies scores → bookkeeping fires contradiction +12 against `LastReporterSupernodeAccount`. Handler then applies explicit +25. Net is +37, but spec §15.1 says +25 only.

## 121-F2 — Recheck +25 / −3 applied to `LastReporterSupernodeAccount` instead of the transcript's reporter
- **Severity**: `HIGH`
- **File**: `x/audit/v1/keeper/msg_storage_truth.go`
- **Lines**: 104-115, 130-141, 152-164, 165-174
- **Status at PR #122 tip**: OPEN — fix is one-line: use `challengedRecord.ReporterAccount` (already loaded line 65) instead of `ticketState.LastReporterSupernodeAccount`.
- **What**: Spec §15.1 references the reporter of the *overturned* fail. Code uses the most recent reporter on the ticket — possibly a different account if a later epoch report wrote to the ticket.

## 121-F3 — Ticket-level contradiction predicate is direction-agnostic — penalises honest reporters
- **Severity**: `HIGH`
- **File**: `x/audit/v1/keeper/storage_truth_scoring.go`
- **Lines**: 575-587, 658-663
- **Status at PR #122 tip**: OPEN at PR #122 (the predicate at lines 718-719 is still bidirectional: `(prev==PASS && currentFailure) || (current==PASS && prevFailure)`).
- **What**: On (prev=PASS, current=fail): the prior honest reporter is marked `contradictedReporter` and gets the −12. Spec §15.1 only penalises the reporter who submitted the bad outcome, AND requires '2 independent clean passes' (or 1 pass + clean recheck) within W=7 — the predicate fires on a single contradicting result.

## 121-F4 — Strong-postpone predicate `LastIndexFailEpoch > 0` fallback never times out
- **Severity**: `HIGH`
- **File**: `x/audit/v1/keeper/enforcement.go`
- **Lines**: 678-679
- **Status at PR #122 tip**: OPEN — guard with `epochDelta(epochID, state.LastIndexFailEpoch) < classAWindow`, or remove the fallback and rely solely on the windowed fact-index.
- **What**: `LastIndexFailEpoch` is set in `updateNodeSuspicionHistoryFields` and never reset. Once a node has any index failure ever, this branch is permanently true. Combined with N transiently spiking ≥140, makes strong-postpone trivially reachable forever.

## 121-F6 — Five new fact indexes (`st/nf/`, `st/rrs/`, `st/spt/`, `st/fh/`, `st/rce/`) never pruned + iteration is full-prefix in DeliverTx and EndBlock
- **Severity**: `HIGH`
- **File**: `x/audit/v1/keeper/prune.go (omission)`
- **Status at PR #122 tip**: OPEN — see also PR #122 Finding 1 (per-result transcript scans in DeliverTx). Add prefix prunes keyed off the embedded u64be epoch_id; ensure `KeepLastEpochEntries >= max(divergence_window, classA_window=14, classB_window=7, oldClassAFaultWindow=21, contradictionWindow=7, patternWindow=14)`.
- **What**: `distinctNodeFailedTickets`, `storageTruthReporterDivergenceStats`, `hasStorageTruthFailedHeal` iterate the entire prefix and filter by epoch in code. Per-result calls in DeliverTx via `MsgSubmitEpochReport`; per-supernode calls every EndBlock at epoch end. Linear gas growth in chain age — canonical Cosmos 'unbounded loop in EndBlock' anti-pattern.

## 121-F7 — `StorageTruthPostponementKey` state has no genesis import/export
- **Severity**: `HIGH`
- **File**: `x/audit/v1/keeper/genesis.go (omission)`
- **Status at PR #122 tip**: OPEN at PR #122 — also include `RecheckEvidence` dedup keys to prevent recheck replay across export round-trip.
- **What**: `setStorageTruthPostponedAtEpochID` writes to `ap/st/<acct>`. ExportGenesis doesn't iterate `StorageTruthPostponementPrefix()`. Chain export+import drops every active storage-truth postponement → instant 'recovery' of every postponed node on restart.

## 121-F8 — `CleanPassCount` is cumulative since first state creation; recovery threshold trivially satisfied
- **Severity**: `HIGH`
- **File**: `x/audit/v1/keeper/storage_truth_scoring.go (~222), x/audit/v1/keeper/enforcement.go (591-595)`
- **Status at PR #122 tip**: OPEN — snapshot `CleanPassCountAtPostpone` when `setStorageTruthPostponedAtEpochID` fires, compare delta only on recovery.
- **What**: Long-tenured node with hundreds of historical PASSes that then accumulates postpone-level N satisfies `CleanPassCount >= requiredPasses` instantly the moment N decays back below threshold — zero post-postponement clean passes required. Spec §17 requires the count *after* postponement.

## 121-F10 — `oldClassAFaultWindow` hardcoded `const = 21`; not parameterised
- **Severity**: `MEDIUM`
- **File**: `x/audit/v1/keeper/enforcement.go:644`
- **Status at PR #122 tip**: PARTIALLY ADDRESSED in PR #122 (added `DefaultStorageTruthOldClassAFaultWindow=21` param). Verify the `const` is removed and the param is used everywhere; otherwise it's a single-source-of-truth violation per invariant-first §3.
- **What**: Window is correct value but not tunable via gov; sibling windows are parameterised.

## 121-F11 — Heal scheduling does not exclude the failing target from healer/verifier candidate pool
- **Severity**: `MEDIUM`
- **File**: `x/audit/v1/keeper/storage_truth_heal_ops.go`
- **Lines**: 248-268, 231-246
- **Status at PR #122 tip**: OPEN — filter active accounts to exclude `state.LastTargetSupernodeAccount`, postponed nodes, and recent class-A offenders for the ticket before passing to `assignStorageTruthHealParticipants`.
- **What**: Spec §18 says deterministic singleton healer, but the entire LEP-6 model exists because the failing holder cannot be trusted. Self-healing the same target you just penalised defeats the point.

## 121-F12 — Strong-postpone band has no behavioural distinction from postpone
- **Severity**: `MEDIUM`
- **File**: `x/audit/v1/keeper/enforcement.go:560-571, 137-152`
- **Status at PR #122 tip**: OPEN — either differentiate (distinct event/reason, stricter recovery: higher `RecoveryCleanPassCount`, longer min-time), or remove the strong-postpone surface to avoid suggesting it does something.
- **What**: `storageTruthBandEventType(strongPostpone)` reuses postpone event; both bands call `setSupernodePostponed` with same reason; recovery treats both identically.

## 121-F13 — `storageTruthEligibleChallengers` ineligibility window check is inverted
- **Severity**: `MEDIUM`
- **File**: `x/audit/v1/keeper/audit_peer_assignment.go:294`
- **Status at PR #122 tip**: OPEN — replace with `if score >= threshold || (state.IneligibleUntilEpoch != 0 && state.IneligibleUntilEpoch >= epochID) { continue }`.
- **What**: `if score >= threshold && (IneligibleUntilEpoch == 0 || IneligibleUntilEpoch >= epochID) { continue }` makes a high-score reporter ineligible whenever `IneligibleUntilEpoch == 0` (i.e. the K=7-epoch ban has never been set). Effectively unbounded ban as soon as score crosses threshold.

## 121-F15 — Divergence engine falls back to stale `WindowPositiveCount/WindowNegativeCount` aggregates
- **Severity**: `MEDIUM`
- **File**: `x/audit/v1/keeper/storage_truth_divergence.go:45-48`
- **Status at PR #122 tip**: OPEN — drop the fallback (use fact-index result authoritatively) or apply same `epochID - WindowStartEpoch >= divergenceWindow` reset at read time.
- **What**: When fact-index returns `stats.total==0`, code falls back to lazy-reset window counts. A long-inactive reporter with stale counts biases the median used by spec §15.2 divergence detection.

## 121-F5 — Old-Class-A predicate fallback uses total `ClassACountWindow` instead of bucket-scoped count
- **Severity**: `MEDIUM`
- **File**: `x/audit/v1/keeper/enforcement.go`
- **Lines**: 651-654
- **Status at PR #122 tip**: OPEN — track `OldClassACountWindow` separately or rely solely on the fact-index path which already enforces bucket+class.
- **What**: Spec §17 requires '2 OLD Class A faults on DISTINCT tickets in 21 epochs'. Code's fallback can be satisfied by 2 RECENT Class A + any single old-bucket fail of any class — over-counts.

## 121-F9 — Class-B postpone fallback uses 14-epoch reset window where spec demands 7
- **Severity**: `MEDIUM`
- **File**: `x/audit/v1/keeper/enforcement.go (639-642), storage_truth_scoring.go (241-245)`
- **Status at PR #122 tip**: OPEN — drop the count-window fallback for Class-B and rely on the windowed fact-index, which already uses `classBWindow=7`.
- **What**: Spec §17: '4 Class B faults in 7 epochs'. `ClassBCountWindow` resets only on `epochID-WindowStartEpoch >= StorageTruthPatternEscalationWindow` (default 14). Fallback over-postpones.

---

This review is posted as a `COMMENT` (not `REQUEST_CHANGES`) so it does not block merge mechanically — but the CRITICAL and HIGH items must be triaged before activation. I'm available to walk through any of these in detail.

— Zee

PR #126, opened 2026-04-28
ci(lint): add diff-only golangci-lint via reviewdog

mateeullahmalik:
## Summary

Adds **diff-only `golangci-lint`** to PRs via [reviewdog](https://github.com/reviewdog/action-golangci-lint). Only NEW lint introduced by the PR's changed lines is reported as inline review comments; the legacy baseline never blocks unrelated PRs.

Pairs with #125 (which clears the BRIDGE-tracked baseline to zero). After both merge, NEW lint that touches `x/action/v1`, `x/supernode/v1`, `x/audit/v1` becomes a hard block on `master`.

## Files

| Path | Purpose |
|---|---|
| `.golangci.yml` | Pinned linter config — `errcheck`, `govet`, `ineffassign`, `staticcheck`, `unused`, `typecheck`. Excludes `*.pb.go` and the SDK-required `HasInvariants` / `InvariantRegistry` SA1019 noise in `x/.../v1/module/module.go`. Both CI and local devs converge on this rule set. |
| `.github/workflows/lint.yml` | `pull_request` workflow running `reviewdog/action-golangci-lint@v2` with `filter_mode: added`, `reporter: github-pr-review`, `fail_level: error`. |

## Behavior

- **On every PR:** reviewdog runs `golangci-lint v1.64.8` against the PR head, computes the diff vs `master`, and posts inline review comments **only on lines the PR added or changed**.
- **`fail_level: error`** → merge is blocked when a new error-level finding lands inside the diff. Pre-existing legacy lint outside the diff is invisible.
- **No third-party account, no billing.** Reviewdog is free for open-source repos and uses the standard `GITHUB_TOKEN`. Required permissions are declared in the workflow (`pull-requests: write`, `checks: write`).

## Local equivalent

```bash
# Diff vs master (matches CI semantics):
golangci-lint run --config=.golangci.yml --new-from-rev=origin/master ./...

# Or full scan of BRIDGE modules (matches the gate report):
golangci-lint run --config=.golangci.yml ./x/action/v1/... ./x/supernode/v1/... ./x/audit/v1/...
```

Devs who want it in their editor: any `golangci-lint` LSP integration (gopls, nvim-lspconfig, VS Code Go) auto-picks `.golangci.yml`.

## Why diff-only

A blanket required `golangci-lint run ./...` on day one would block every PR until the full repo is cleaned. Diff-only avoids that — legacy noise stays invisible until a PR touches that line, at which point it becomes the toucher's problem to fix or `//nolint` with a reason. Allows incremental cleanup with zero velocity loss.

## Rollout plan after this lands

1. Merge #125 (clears BRIDGE-module baseline to 0).
2. Merge this PR. Diff-only enforcement starts immediately.
3. Watch for ~2 weeks. If false-positive rate is acceptable, expand by:
   - Adding more linters (e.g. `gosimple`, `gocritic`, `revive`) one-at-a-time.
   - Tightening `staticcheck` checks list.
4. Optional later: add a separate weekly **full-repo** lint job that posts a Datadog metric of total diagnostic count, aiming for monotonic decrease.

## Risks

- Minimal. Diff-only mode means the workflow cannot regress merging on legacy noise.
- Reviewdog is well-maintained (used by Kubernetes, Prometheus, gRPC, ~5k stars). v2 is the current major; we pin the minor by default but accept the v2 rolling tag for security patches — if you'd rather pin a SHA, say so and I'll update.
- Worst case if the action breaks: lint job fails, but `test.yml` and other required checks still gate merge. Removing or disabling the workflow is a one-line change.

## Verification

I ran the same config locally against the current `master` baseline (pre-#125): produces the same 25 findings the gate report listed, exit 0 (because `fail_level` only triggers on diff-introduced issues). After #125 merges, baseline goes to 0.

## Note on cost

Reviewdog and golangci-lint are both **free** for any GitHub repo, public or private. The only resource cost is GitHub Actions minutes — public repos get unlimited free minutes; private orgs pay the standard Actions runner rate. Job runtime on this repo is ~2 minutes.


mateeullahmalik:
Reworked into an **additive** PR-only layer after rebasing on master (which now ships its own full-scan `golangci-lint` hard gate). Net diff against master is **one new file**: `.github/workflows/lint-pr.yml` (reviewdog diff-only + govulncheck). Master's `lint.yml` and `.golangci.yml` are left untouched.

Copilot review applied:

| Comment | Fix |
|---|---|
| `level: warning` neuters `fail_level: error` | Removed `level`; native severities preserved so error-level diagnostics actually trip the gate. |
| Fork PRs can't post review comments | Step split on `head.repo.full_name == github.repository`. Same-repo PRs → `github-pr-review` (inline). Fork PRs → `reporter: local` (log-only). |
| `typecheck` claimed enabled in description but isn't | Description rewritten; no longer claims `typecheck`. |
| `go.mod` pins golangci-lint v1.64.8 → `make install-tools` breaks v2 schema parsing | Documented in PR description and added install command for v2.11.4 out of band. (Not bumping the go.mod tool path here — it cascades into ~25 unrelated dep upgrades. Tracked separately.) |
| PR description vs workflow version mismatch | Description rewritten; reflects `v2.11.4` everywhere. |
| (Original report) govulncheck job failure | Root cause was duplicate `Authorization` header from the action's internal `actions/checkout`. Fixed with `repo-checkout: false`. |

j-rafique (approved):

PR #152, opened 2026-06-16
scripts(evmigration): make migrate-multisig.sh submit recoverable on broadcast RPC drop

mateeullahmalik:
## Summary

Make `scripts/migrate-multisig.sh submit` recoverable when the CometBFT RPC stream drops mid-broadcast (lumerad exits non-zero after printing `EOF` to stderr). Today the wrapper dies under `set -e` with no diagnostic and a bare `EOF` line — the operator has no way to know whether the tx actually made it into the mempool.

## Why

Observed in the wild on `lumera-devnet-1` during the `val2` multisig validator migration ceremony. The `migrate-multisig.sh submit` wrapper printed:

```
gas estimate: 6871045
EOF
```

…and exited. But the tx had in fact landed cleanly at the next block (`migrate_validator` event, `code: 0`, h=51008). The operator's only recourse without this change is to manually query the migration record themselves and figure out whether the tx landed. If they guess wrong:

- Tx **did** land + operator re-submits → second attempt fails with `migration record already exists` (confusing but harmless).
- Tx **did not** land + operator re-submits → exactly the right move, but only if you're sure.

There's no way for the operator to be sure from the wrapper's output alone.

## What this PR changes

`scripts/migrate-multisig.sh` — `_mms_submit`:

- Capture `lumerad tx evmigration submit-proof` stdout/stderr/exit code separately instead of letting `set -e` kill the script.
- On non-zero exit:
  - Log the lumerad stderr (so the operator sees `EOF` or whatever the real error was, not just a silent bash exit).
  - Query the on-chain migration record for the legacy address.
  - If the record exists and `new_address` matches the destination from `tx.json`, log a recovery message, run the standard `verify_migration` + `show_migration_summary` path, and exit 0.
  - If no record exists, log `tx did not land; safe to re-run submit` and exit 2 (existing broadcast-rejected sentinel).

`tests/scripts/fixtures/lumerad-shim.sh`:

- New env vars `SHIM_SUBMIT_EXIT`, `SHIM_SUBMIT_STDERR`, `SHIM_SUBMIT_LANDED` so tests can simulate "broadcast bombs with EOF but tx landed" and "broadcast bombs and tx didn't land".

`tests/scripts/migrate-multisig.bats`:

- `submit recovers when broadcast prints EOF but tx actually landed` — asserts exit 0 + recovery log + verification ran.
- `submit exits 2 with clear retry guidance when broadcast fails and tx did not land` — asserts exit 2 + retry guidance in the output.

## Scope

- Only `migrate-multisig.sh submit`. `migrate-account.sh` and `migrate-validator.sh` are single-signer and don't expose the same operator-confusion failure mode (a single-key broadcast that returns EOF is just a retry — there's no offline ceremony in front of it to worry about discarding).
- No protocol or chain-state changes.
- No new dependencies (uses `mktemp`, `lumerad_q`, `jq` — already required by the script).

## Test plan

```
$ bats tests/scripts/migrate-multisig.bats
ok 1..43

$ bats tests/scripts/
1..121
… all 121 tests pass
```

Locally rehearsed by hand: pointed the script at a node whose RPC was getting killed mid-flight and confirmed both branches (recovery / retry-guidance) trigger as expected.

## Risks

Low. The change is fenced inside the `(( broadcast_rc != 0 ))` branch — happy-path behavior is unchanged (existing `submit happy path (broadcast + verify) exits 0` test still passes).

## Rollback strategy

Revert this PR. No state-machine changes, no migration record on-chain, no schema changes.

### DIFF
--- a/scripts/migrate-multisig.sh
+++ b/scripts/migrate-multisig.sh
@@ -631,8 +631,38 @@ BANNER
   [[ -n "$home_dir"    ]] && args+=(--home "$home_dir")
   (( yes == 1 )) && args+=(-y)
 
-  local broadcast_json tx_hash
-  broadcast_json=$("$BIN" "${args[@]}")
+  # Capture stdout, stderr, and exit code separately so a partial RPC failure
+  # (e.g. lumerad prints `EOF` to stderr after the CometBFT RPC stream drops)
+  # doesn't kill the script under `set -e` with no diagnostic. The tx may or
+  # may not have entered the mempool; before erroring out, probe chain state
+  # for the migration record so the operator gets a definitive answer instead
+  # of a bare `EOF` line.
+  local broadcast_json broadcast_err broadcast_rc=0 tx_hash
+  local stderr_file
+  stderr_file=$(mktemp)
+  broadcast_json=$("$BIN" "${args[@]}" 2>"$stderr_file") || broadcast_rc=$?
+  broadcast_err=$(<"$stderr_file")
+  rm -f "$stderr_file"
+  if (( broadcast_rc != 0 )); then
+    log_error "broadcast command failed (rc=$broadcast_rc)"
+    [[ -n "$broadcast_err" ]] && log_error "lumerad stderr: $broadcast_err"
+    log_error "checking chain state — the tx may still have landed:"
+    log_error "  $BIN query evmigration migration-record $(legacy_value "$legacy") --node $node"
+    if lumerad_q evmigration migration-record "$legacy" >/dev/null 2>&1; then
+      local rec_new
+      rec_new=$(lumerad_q evmigration migration-record "$legacy" 2>/dev/null | jq -r '.record.new_address // empty')
+      if [[ "$rec_new" == "$new" ]]; then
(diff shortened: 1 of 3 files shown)


akobrin1:
**P2: recovery path can falsely report that a pending tx did not land**

In `scripts/migrate-multisig.sh`, the non-zero `submit-proof` recovery branch queries `migration-record` immediately and then prints `tx did not land; safe to re-run submit` when no record exists. Because `submit-proof` uses sync broadcast, an RPC EOF can happen after mempool acceptance but before the next block creates the migration record. In that window this path gives a false negative and may tell the operator to retry while the original tx is still pending.

Could we poll for the migration record until `LUMERA_TX_WAIT_TIMEOUT` or a short block-based timeout before concluding it did not land? It would also be good to add a bats case where the record appears only after one or more failed/empty recovery queries.

Reference: `scripts/migrate-multisig.sh`, around the new recovery branch at lines 646-664.

a-ok123 (dismissed):

j-rafique (approved):

PR #192, opened 2026-07-14
evmigration: repair v1.20.0 raw-shares DelegatorStartingInfo at migration time

mateeullahmalik:
## Summary

Repairs the wedged `DelegatorStartingInfo` rows written by Lumera **v1.20.0**'s `migrateActiveDelegations`, which stored `Stake = del.Shares` instead of `TokensFromSharesTruncated(del.Shares)`. On any validator that was ever slashed, `WithdrawDelegationRewards` for those delegators now panics inside the SDK sanity check:

```
calculated final stake for delegator … greater than current stake
```

That panic bricks `tx evmigration migrate-validator` (and any code path that withdraws for these delegators), which is how Alexy first hit it on testnet-2.

**No chain upgrade is required.** The repair is in-flight: called immediately before every `WithdrawDelegationRewards` inside `x/evmigration`.

Core: The SDK sanity check trips because the *stored* starting-info stake is raw shares (pre-slash-like), not tokens, so `IterateValidatorSlashEventsBetween` can't reconcile the difference against a post-slash validator pool. Verified against hotfix commit `4ce27cf0` ("store delegation starting stake as tokens rather than shares").

## Root cause 

- v1.20.0 `migrateActiveDelegations` wrote `DelegatorStartingInfo{Stake: del.Shares, …}`.
- Cosmos SDK v0.53.x invariant in `x/distribution/keeper/delegation.go` (calculateDelegationRewards): `stake ≤ currentStake`. For a slashed validator, raw shares > tokens, so the check trips and the tx panics.
- v1.20.1 hotfix (`4ce27cf0`, 2026-07-06) fixed the *writer* to use `TokensFromSharesTruncated`. Existing rows written under v1.20.0 remain wedged.
- Fix must repair rows in-place at read time; a chain upgrade / governance handler is not required.

## Blast radius (measured on live state)

Scanned every `DelegatorStartingInfo` row on testnet-2 (h=5,665,862) and mainnet-1 (h=5,961,728) for the shape `stored.Stake > val.TokensFromSharesTruncated(shares)`.

**testnet-2 — 5 wedged pairs:**

| Delegator | Validator | Shares | Stored `Stake` (wrong) | Expected after repair | Migration tx height |
|---|---|---|---|---|---|
| `lumera1w3phq…` | `…1rq08uj8` (Nodeist) | 0.00000047 | 0.000000469851 | 0.000000468912 | 5552035 |
| `lumera1w3phq…` | `…1ux9pyka` (light-val-1) | 500000 | 500000.0 | 499000.500611 | 5552035 |
| `lumera15vdx…` | `…1qh38y670` (ADRnode) | 500 | 500.0 | 499.5426 | 5557495 |
| `lumera15vdx…` | `…18r2hdxu3` (Fertal) | 502.005 | 502.005 | 499.0005 | 5557495 |
| `lumera1vny5z…` | `…1gtwgx5d3` (Nodeyus) | 10020.03 | 10020.03 | 10000.0 | 5574049 |

All three delegators' `DelegatorStartingInfo.Height` fields match their respective `MsgClaimLegacyAccount` tx heights, all under v1.20.0.

**mainnet-1 — 0 wedged pairs** across 2,298 delegators on 25 slashed validators. This is a testnet-only cleanup, but the repair remains valid indefinitely as a defensive layer.

## Change

New helper — `x/evmigration/keeper/repair_distribution_stake.go`:

```go
func (k Keeper) RepairV120DistributionStake(
    ctx sdk.Context, valAddr sdk.ValAddress, delAddr sdk.AccAddress,
    del stakingtypes.Delegation,
) (repaired bool, err error)
```

Reads current `DelegatorStartingInfo`, computes `expected = val.TokensFromSharesTruncated(del.Shares)`, only writes when `stored.Stake > expected`. Preserves `PreviousPeriod` and `Height`. Semantically equivalent to running the v1.20.1 writer retroactively. Emits `evmigration_v120_stake_repair` event with `{delegator, validator, old_stake, new_stake, shares}`. Returns typed `ErrDistributionStateInconsistent` if `stored.Stake < expected` (impossible-shape guard — never observed, but a safety net).

**Two call sites — the only two `WithdrawDelegationRewards` in `x/evmigration`:**

1. `keeper/migrate_distribution.go:38` (per-delegation loop in `MigrateDistribution`)
2. `keeper/msg_server_migrate_validator.go:166` (per-delegator loop in `MigrateValidator`)

`MigrationEstimate` (query.go) extended with a new `distribution_repairs_needed` field on `QueryMigrationEstimateResponse` so operators can preview affected rows before submitting.

## Invariants preserved

| # | Invariant | How preserved |
|---|---|---|
| I1 | Repair writes tokens ≡ output of v1.20.1 writer | `expected = TokensFromSharesTruncated(shares)` |
| I2 | Only touches the exact v1.20.0 bug signature | `stored.Stake > expected` gate; never widens |
| I3 | Idempotent | Second call is a no-op (already ≤ expected) |
| I4 | Deterministic | No wall-clock, no map iteration, only KV reads |
| I5 | Observable | Event emission on every mutation |
| I6 | Surfaced in preview | `MigrationEstimate` returns per-pair repair rows |

## Tests

- **6 direct-helper unit tests** (`repair_distribution_stake_test.go`): wedge repaired + event, healthy no-op, `stored < expected` typed error, missing row no-op, non-existent validator error, idempotent re-run.
- **7 existing tests updated** with matching mocks (`expectHealthyStakeRepair` helper for the no-op path).
- **All `x/` unit tests pass.**
- **Integration tests pass:** `go test -tags='integration test' ./tests/integration/evmigration/... -count=1` → `ok 10.779s`.
- `go vet` clean, `make lint` reports 0 issues.

## Rollback

Revert this commit; behavior returns to panic-on-wedged-delegator. No state migration to unwind — the repair only rewrites the exact rows the v1.20.1 hotfix would have written. Wedged rows previously repaired remain valid under both v1.20.0 and v1.20.1+ readers.

## Determinism / gas / upgrade / IBC

- Deterministic: pure KV reads + one conditional KV write per wedged pair.
- Gas: bounded by one `GetValidator` + one `GetDelegatorStartingInfo` + at most one `SetDelegatorStartingInfo` per delegator processed. `MigrationEstimate` iteration is unchanged.
- Upgrade: none required.
- IBC: N/A.

## Follow-ups (out of scope for this PR)

- Keyring UX bug in `migrate-validator.sh` (Alexy's issue #1) — separate script fix.
- Preflight/UX surfacing of `distribution_repairs_needed` in the CLI wrapper.

/cc Kay Alexy ak


akobrin1 (commented):
PR #191 (merged) already provided the state repair + regression tests + documentation.
In #192 - different repair algorithm that is less precise:

- It treats any startingInfo.Stake > current TokensFromShares(...) as corruption.
That condition is legitimate when a slash occurred after the starting-info row; Cosmos SDK expects to replay that slash.
- It replaces the starting stake with the current token stake while preserving the old height and period, so later slash replay can apply the slash twice and understate rewards. 
- That addition (distribution_repairs_needed in MigrationEstimate) is specific to testnet only (the issue won't happen in mainnet - so the option will just confuse mainnet users) - don't think we need it.
The additional MigrateDistribution call site is probably unnecessary for this historical bug because corrupted rows belong to accounts that already completed migration under v1.20.0.

mateeullahmalik:
superseded

